PR #1, opened 2015-02-19
Initial MySQL repository

mnagy:
My first take on OpenShift MySQL image.

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -1,0 +1,35 @@
+FROM centos:centos7
+
+# MySQL image for OpenShift.
+#
+# Volumes:
+#  * /var/lib/mysql - Datastore for MySQL
+# Environment:
+#  * $MYSQL_USER - Database user name
+#  * $MYSQL_PASSWORD - User's password
+#  * $MYSQL_DATABASE - Name of the database to create
+
+# Image metadata
+ENV MYSQL_VERSION           = "5.5" \
+    IMAGE_DESCRIPTION       = "MySQL 5.5" \
+    IMAGE_TAGS              = "mysql,mysql55" \
+    IMAGE_EXPOSE_SERVICES   = "3306:mysql"
+
+MAINTAINER  Martin Nagy <mnagy@redhat.com>
+EXPOSE 3306
+
+RUN groupadd -g 27 -o -r mysql && \
+    useradd -M -N -g mysql -o -r -d /var/lib/mysql -s /sbin/nologin -c "MySQL Server" -u 27 mysql
+
+RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
+    yum -y --setopt=tsflags=nodocs install https://www.softwarecollections.org/en/scls/rhscl/mysql55/epel-7-x86_64/download/rhscl-mysql55-epel-7-x86_64.noarch.rpm && \
+    yum -y --setopt=tsflags=nodocs install hostname mysql55 && \
+    yum clean all
+
+COPY mysql /opt/openshift/
+ENV PATH /opt/openshift/bin:$PATH
+WORKDIR /opt/openshift
+
+VOLUME ["/var/lib/mysql"]
+
+ENTRYPOINT ["run-mysqld.sh"]
--- a/5.5/Dockerfile.rhel7
+++ b/5.5/Dockerfile.rhel7
@@ -1,0 +1,36 @@
+FROM rhel7
+
+# MySQL image for OpenShift.
+#
+# Volumes:
+#  * /var/lib/mysql - Datastore for MySQL
+# Environment:
+#  * $MYSQL_USER - Database user name
+#  * $MYSQL_PASSWORD - User's password
+#  * $MYSQL_DATABASE - Name of the database to create
+
+# Image metadata
+ENV MYSQL_VERSION           = "5.5" \
+    IMAGE_DESCRIPTION       = "MySQL 5.5" \
+    IMAGE_TAGS              = "mysql,mysql55" \
+    IMAGE_EXPOSE_SERVICES   = "3306:mysql"
+
+MAINTAINER  Martin Nagy <mnagy@redhat.com>
+EXPOSE 3306
+
+RUN groupadd -g 27 -o -r mysql && \
+    useradd -M -N -g mysql -o -r -d /var/lib/mysql -s /sbin/nologin -c "MySQL Server" -u 27 mysql
+
+RUN yum install -y yum-utils hostname && \
+    yum-config-manager --enable rhel-server-rhscl-7-rpms && \
+    yum-config-manager --enable rhel-7-server-optional-rpms && \
+    yum install -y --setopt=tsflags=nodocs mysql55 && \
+    yum clean all
+
+COPY mysql /opt/openshift/
+ENV PATH /opt/openshift/bin:$PATH
+WORKDIR /opt/openshift
+
+VOLUME ["/var/lib/mysql"]
+
+ENTRYPOINT ["run-mysqld.sh"]
--- a/5.5/mysql/etc/my.cnf
+++ b/5.5/mysql/etc/my.cnf
@@ -1,0 +1,13 @@
+[mysqld]
+user=mysql
+
+datadir=/var/lib/mysql/data
+basedir=/opt/rh/mysql55/root/usr
+plugin-dir=/opt/rh/mysql55/root/usr/lib64/mysql/plugin
+
+general_log=ON
+
+socket=/var/lib/mysql/mysql.sock
+
+# Disabling symbolic-links is recommended to prevent assorted security risks
+symbolic-links=0
--- a/README.md
+++ b/README.md
@@ -1,0 +1,26 @@
+# OpenShift MySQL image
+
+This repository contains Dockerfiles for MySQL images for OpenShift. Users can choose between RHEL and CentOS based images.
+
+
+# Installation and Usage
+
+Choose between CentOS7 or RHEL7 base image:
+
+*  **RHEL7 base image**
+
+To build a base-rhel7 image, you need to run docker build it on properly subscribed RHEL machine.
+
+```
+$ git clone https://github.com/openshift/mysql.git
+$ cd mysql
+$ make build TARGET=rhel7
+```
+
+*  **CentOS7 base image**
+
+```
+$ git clone https://github.com/openshift/mysql.git
+$ cd mysql
+$ make build
+```
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -1,0 +1,40 @@
+#!/bin/bash -e
+# $1 - Specifies distribution - RHEL7/CentOS7
+# $2 - Specifies MySQL version - 5.5
+
+# Array of all versions of MySQL
+declare -a VERSIONS=(5.5)
+
+OS=$1
+VERSION=$2
+
+
+function build_rhel {
+  mv Dockerfile Dockerfile.centos7
+  mv Dockerfile.rhel7 Dockerfile
+  trap "mv Dockerfile Dockerfile.rhel7 && mv Dockerfile.centos7 Dockerfile" RETURN
+  docker build -t ${IMAGE_NAME} .
+}
+
+if [ -z ${VERSION} ]; then
+  # Build all versions
+  dirs=${VERSIONS}
+else
+  # Build only specified version on MySQL
+  dirs=${VERSION}
+fi
+
+for dir in ${dirs}; do
+  IMAGE_NAME=openshift/mysql-${dir//./}-${OS}
+  echo ">>>> Building ${IMAGE_NAME}"
+
+  pushd ${dir} > /dev/null
+
+  if [ "$OS" == "rhel7" ]; then
+    build_rhel
+  else
+    docker build -t ${IMAGE_NAME} .
+  fi
+
+  popd > /dev/null
+done
--- a/5.5/mysql/bin/run-mysqld.sh
+++ b/5.5/mysql/bin/run-mysqld.sh
@@ -1,0 +1,43 @@
+#!/bin/bash -e
+
+function usage {
+	echo "You must specify following environment variables:"
+	echo "  \$MYSQL_USER"
+	echo "  \$MYSQL_PASSWORD"
+	echo "  \$MYSQL_DATABASE"
+	exit 1
+}
+
+test -z "$MYSQL_USER" && usage
+test -z "$MYSQL_PASSWORD" && usage
+test -z "$MYSQL_DATABASE" && usage
+
+if [ ! -d '/var/lib/mysql/data' ]; then
+
+	echo 'Running mysql_install_db ...'
+	scl enable mysql55 "mysql_install_db --user=mysql --datadir=/var/lib/mysql/data"
+	echo 'Finished mysql_install_db'
+
+	# These statements _must_ be on individual lines, and _must_ end with
+	# semicolons (no line breaks or comments are permitted).
+	# TODO proper SQL escaping on ALL the things D:
+	TEMP_FILE='/tmp/mysql-first-time.sql'
+	cat > "$TEMP_FILE" <<-EOSQL
+		DELETE FROM mysql.user ;
+		CREATE USER '${MYSQL_USER}'@'%' IDENTIFIED BY '${MYSQL_PASSWORD}' ;
+		GRANT ALL ON *.* TO '${MYSQL_USER}'@'%' WITH GRANT OPTION ;
+		DROP DATABASE IF EXISTS test ;
+		CREATE DATABASE IF NOT EXISTS \`${MYSQL_DATABASE}\` ;
+		FLUSH PRIVILEGES ;
+	EOSQL
+
+	set -- "$@" --init-file="$TEMP_FILE"
+fi
+
+chown -R mysql:mysql /var/lib/mysql
+
+# SCL in CentOS/RHEL 7 doesn't support --exec, we need to do it ourselves
+export X_SCLS="mysql55"
+source /opt/rh/mysql55/enable
+
+exec /opt/rh/mysql55/root/usr/libexec/mysqld --defaults-file=/opt/openshift/etc/my.cnf "$@" 2>&1
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,17 @@
+
+ifeq ($(TARGET),rhel7)
+	OS := rhel7
+else
+	OS := centos7
+endif
+
+ifeq ($(VERSION), 5.5)
+	VERSION := 5.5
+else
+	VERSION :=
+endif
+
+.PHONY: build
+build:
+	hack/build.sh $(OS) $(VERSION)
+


5.5/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM centos:centos7
+
```
mfojtik:
```
MAINTAINER  Martin Nagy <mnagy@redhat.com>
```

5.5/Dockerfile.rhel7
```diff
@@ -0,0 +1,20 @@
+FROM rhel7
+
```
mfojtik: same here

5.5/mysql/bin/docker-entrypoint.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
```
mfojtik: add -e here and nuke the line below

5.5/mysql/bin/docker-entrypoint.sh
```diff
@@ -0,0 +1,47 @@
...
+set -e
+
+# TODO: Only allow mysqld_safe?
```
mfojtik: yes. you can remove this comment now.

5.5/mysql/bin/docker-entrypoint.sh
```diff
@@ -0,0 +1,47 @@
...
+if [ ! -d '/var/lib/mysql/mysql' -a "${1%_safe}" = 'mysqld' ]; then
+	if [ -z "$MYSQL_ROOT_PASSWORD" ]; then
```
mfojtik: It should print the usage if these vars are not set (instead of starting the mysql server with wrong parameters)

5.5/mysql/bin/docker-entrypoint.sh
```diff
@@ -0,0 +1,47 @@
...
+	
+	scl enable mysql55 -- mysql_install_db --user=mysql --datadir=/var/lib/mysql
+	
+	# These statements _must_ be on individual lines, and _must_ end with
+	# semicolons (no line breaks or comments are permitted).
+	# TODO proper SQL escaping on ALL the things D:
+	TEMP_FILE='/tmp/mysql-first-time.sql'
+	cat > "$TEMP_FILE" <<-EOSQL
+		DELETE FROM mysql.user ;
+		CREATE USER 'root'@'%' IDENTIFIED BY '${MYSQL_ROOT_PASSWORD}' ;
```
mfojtik: this should be MYSQL_ROOT_USER ?

5.5/mysql/bin/docker-entrypoint.sh
```diff
@@ -0,0 +1,47 @@
...
+		GRANT ALL ON *.* TO 'root'@'%' WITH GRANT OPTION ;
+		DROP DATABASE IF EXISTS test ;
+	EOSQL
+	
+	if [ "$MYSQL_DATABASE" ]; then
```
mfojtik: So if this is not set, we don't create the dabase? How is then the mysql image usefull?

5.5/mysql/bin/docker-entrypoint.sh
```diff
@@ -0,0 +1,47 @@
...
+	fi
+	
+	echo 'FLUSH PRIVILEGES ;' >> "$TEMP_FILE"
+	
+	set -- "$@" --init-file="$TEMP_FILE"
+fi
+
+chown -R mysql:mysql /var/lib/mysql
+
+export X_SCLS="mysql55"
```
mfojtik: do this at top and you don't need to call scl enable at line 12

5.5/mysql/etc/my.cnf
```diff
@@ -0,0 +1,7 @@
+[mysqld]
+datadir=/var/lib/mysql
+# Disabling symbolic-links is recommended to prevent assorted security risks
+symbolic-links=0
+
+#[mysqld_safe]
+#pid-file=<%= ENV['OPENSHIFT_MYSQL_DIR'] %>pid/mysql.pid
```
mfojtik: there will be no PID file. the container itself is the process and it has pid

README.md
```diff
@@ -0,0 +1,28 @@
+# OpenShift MySQL image
+
+This repository contains dockerfiles for MySQL images for OpenShift. Users can choose between RHEL and CentOS based images.
```
mfojtik: Dockerfiles

README.md
```diff
@@ -0,0 +1,28 @@
...
+
+
+# Installation and Usage
+
+Choose between CentOS7 or RHEL7 base image:
+
+*  **RHEL7 base image**
+
+This image is not available as trusted build in [Docker Index](https://index.docker.io).
```
mfojtik: We will not have trusted build for RHEL image.

mfojtik:
@bparees @soltysh PTAL ^^ 

this is our first database image, we should do it right ;-)

mfojtik:
@mnagy I would like to see what is your testing plan for this image.

5.5/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+ADD mysql /opt/openshift/
+ENV PATH /opt/openshift/bin:$PATH
+
+WORKDIR /opt/openshift
+RUN create_user.sh
```
soltysh: I don't see any sense in having separate script for that, do this here, like we do in [sti-nodejs](https://github.com/openshift/sti-nodejs/blob/master/0.10/Dockerfile#L16-L19).

5.5/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
+    yum -y --setopt=tsflags=nodocs install https://www.softwarecollections.org/en/scls/rhscl/mysql55/epel-7-x86_64/download/rhscl-mysql55-epel-7-x86_64.noarch.rpm && \
+    yum -y --setopt=tsflags=nodocs install hostname mysql55 && \
+    yum clean all
+
+
+VOLUME ["/var/lib/mysql"]
+
+ENTRYPOINT ["entrypoint.sh"]
```
soltysh: You're calling `entrypoint.sh` whereas the file is called `docker-entrypoint.sh`.

5.5/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+EXPOSE 3306
+CMD ["mysqld_safe", "--defaults-file=./etc/my.cnf"]
```
soltysh: Why having both entrypoint and cmd, the first should suffice.

README.md
ncdc: s/trusted/automated

5.5/Dockerfile
bparees: Needs descriptive comments in general at the top.

5.5/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+    yum -y --setopt=tsflags=nodocs install https://www.softwarecollections.org/en/scls/rhscl/mysql55/epel-7-x86_64/download/rhscl-mysql55-epel-7-x86_64.noarch.rpm && \
+    yum -y --setopt=tsflags=nodocs install hostname mysql55 && \
+    yum clean all
+
+
+VOLUME ["/var/lib/mysql"]
+
+ENTRYPOINT ["entrypoint.sh"]
+
+EXPOSE 3306
```
bparees: @mfojtik can you help @mnagy put in the other metadata you've defined for images?

5.5/Dockerfile
bparees: just embed the commands directly here, i'm not seeing much value in doing it this way, plus you end up with the script in your image at the end.

5.5/Dockerfile.rhel7
```diff
@@ -0,0 +1,20 @@
...
+    yum-config-manager --enable rhel-server-rhscl-7-rpms && \
+    yum-config-manager --enable rhel-7-server-optional-rpms && \
+    yum install -y --setopt=tsflags=nodocs mysql55 && \
+    yum clean all
+
+VOLUME ["/var/lib/mysql"]
+
+ENTRYPOINT ["entrypoint.sh"]
+
+EXPOSE 3306
```
bparees: move this up near the ENV content

5.5/mysql/bin/docker-entrypoint.sh
bparees: yeah this should probably be mandatory along w/ the root and normal passwords and the user name.

bparees: I think it would be ok for the image to provide defaults for these values in the Dockerfile, too.

bparees: or not...we recommend against default passwords here https://ci.openshift.redhat.com/openshift-docs-master-testing/latest/image_writers_guide/guidelines.html  (see "avoid default passwords")
can probably default the other bits though.

hack/build.sh
```diff
@@ -0,0 +1,40 @@
...
+declare -a VERSIONS=(5.5)
+
+OS=$1
+VERSION=$2
+
+
+function build_rhel {
+  mv Dockerfile Dockerfile.centos7
+  mv Dockerfile.rhel7 Dockerfile
+  trap "mv Dockerfile Dockerfile.rhel7 && mv Dockerfile.centos7 Dockerfile" RETURN
```
bparees: good idea, can you make this change in our other STI repos?  it would be good to keep this logic identical across our repos.

5.5/Dockerfile.rhel7
```diff
@@ -0,0 +1,20 @@
...
+
+RUN yum install -y yum-utils hostname && \
+    yum-config-manager --enable rhel-server-rhscl-7-rpms && \
+    yum-config-manager --enable rhel-7-server-optional-rpms && \
+    yum install -y --setopt=tsflags=nodocs mysql55 && \
+    yum clean all
+
+VOLUME ["/var/lib/mysql"]
+
+ENTRYPOINT ["entrypoint.sh"]
```
bparees: docker-entrypoint.sh?  

might make more sense to just call it "run" to align w/ our STI image run scripts.

bparees:
can you think of a way to test this in a way similar to what we do in the STI repos?  obviously there's less you can do, but at least starting the image and curl'ing the mysql port and confirming the response string (it's mostly garbage but i think there's some text in their you can parse)

better would be a real client that could connect and confirm a DB existed w/ the right password.  I'm not sure i want to drag DB drivers and code into this repo just for testing, but it might be a good idea.

hack/build.sh
jhadvig: +1

5.5/Dockerfile
mfojtik: @bparees +1 we should put as many comments into Dockerfile(s) as we can. Especially to describe our reasoning of why we did the things in this way, to avoid future confusion.

5.5/mysql/bin/docker-entrypoint.sh
mfojtik: @bparees operations will hate to have default passwords... I would recommend to have these as requirements and if they are not set, then the image should error out (and print usage).

README.md
mfojtik: @ncdc automated trusted build ;D

5.5/Dockerfile
mfojtik: That was my idea actually... I found having this in Dockerfile pretty ugly and hard to read. Since he is adding the 'mysql' directory in 2 steps before, the create_user.sh addition is noop here.

Also in future, we can put more logic into user creation, etc. Maybe we should make the script accepting parameter (username?).

The script inside the image is nice in my opinion because you can easily inspect how the user was created and with what parameters.

5.5/Dockerfile
mfojtik: @soltysh it is what upstream mysql image is doing and I kinda like that. I was confused myself when I saw it for the first time. In this way, you always make sure that the entrypoint (or whatever this script will be called) handle pre-evaluation of the database creation and set the right env vars. The CMD is then explicit and you can see what is actually being run.

hack/build.sh
mfojtik: @bparees and here we are, copying and pasting code across repos :-D

5.5/Dockerfile.rhel7
mfojtik: maybe be more descriptive and call it `run-scl-mysql.sh`

mfojtik:
@bparees we also need to test setting the MYSQL_ env vars and making sure they are picked up. To test this image, the `test/run` script should spin up the container, set the ENV vars and then call the mysql client and connect into the container (similar to HTTP check we do in STI images). Then execute some SQL query (SELECT 1?). The nice thing of having the ENTRYPOINT bootstrapping the SCL is that you can call 'mysql' binary from this image, so it does not need to exists in the system you are testing this image...

```
docker run openshift/mysql-55-centos7 mysql -u MYSQL_USER -p MYSQL_PASSWORD "SELECT 1"
```

Then check the return code.

mfojtik:
@mnagy sorry for closing this PR, I hit wrong button :(

5.5/Dockerfile
mnagy: I tend to agree with @mfojtik, having it in Dockerfile is ugly, IMO. I agree that script ending up in the image is not ideal, but it's just around 135 bytes, so I don't think we need to cut corners there.

I would add arguments for UID, GID and name.

5.5/Dockerfile
mnagy: Any suggestions?

5.5/Dockerfile.rhel7
mnagy: entrypoint.sh is from upstream mysql image. While `run-scl-mysql.sh` is more descriptive, I dislike dropping "entrypoint" from the name, because that makes it clear that the script serves as an entrypoint. Maybe `entrypoint-scl-mysql.sh`?

5.5/Dockerfile
mfojtik: @mnagy just say what is this image, why it exists, what it require to run

5.5/Dockerfile
mfojtik: @mnagy UID and GID would be nice to tell as well

5.5/Dockerfile.rhel7
mfojtik: @mnagy ENTRYPOINT instruction is already there that marks it as entrypoint... no need for duplications imho.

5.5/mysql/bin/docker-entrypoint.sh
soltysh: +1

README.md
mnagy: Will remove the entire line after discussion with Michal off-github

5.5/Dockerfile.rhel7
mnagy: Why? After some thought, I think it's best right after entrypoint. why have intermediate images with exposed port when there is no chance of it being used?

mnagy: Official mysql image is doing that as well.

mfojtik: @mnagy EXPOSE does not provide any functionality, it just set the metadata for the image. We are not going to provide the intermediate images anyway. Also the EXPOSE will most likely not change during the time, so I think what @bparees suggests is correct.

5.5/Dockerfile
bparees: @mnagy it's not about the size, it's about the clutter.  it also makes it harder to read the Dockerfile because i need to go jump off to some other script to know what you're doing in it.  I've never seen a Docker image that does this, can you show me one?

5.5/mysql/bin/docker-entrypoint.sh
bparees: agree on passwords, we should still default the others.

5.5/Dockerfile.rhel7
bparees: my reasoning is twofold:
1) it's not likely to change so it should go earlier
2) i'm trying to group the "metadata" type stuff in one location for convenience when people read the dockerfile.

mnagy:
Everything except testing should be addressed, PTAL.

soltysh:
@mnagy please squash all the changes before merging this one.

5.5/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+RUN groupadd -g 27 -o -r mysql && \
+    useradd -M -N -g mysql -o -r -d /var/lib/mysql -s /sbin/nologin -c "MySQL Server" -u 27 mysql
+
+RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
+    yum -y --setopt=tsflags=nodocs install https://www.softwarecollections.org/en/scls/rhscl/mysql55/epel-7-x86_64/download/rhscl-mysql55-epel-7-x86_64.noarch.rpm && \
+    yum -y --setopt=tsflags=nodocs install hostname mysql55 && \
+    yum clean all
+
+ADD mysql /opt/openshift/
```
mfojtik: we use COPY for this in other images, maybe be consistent? https://github.com/mfojtik/sti-nodejs/blob/metadata/0.10/Dockerfile#L20

5.5/mysql/bin/entrypoint.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash -e
+
+function usage {
+	echo "You must specify following environment variables:"
+	echo "  \$MYSQL_ROOT_PASSWORD"
```
mfojtik: Why you need 2 passwords? (ROOT_PASSWORD && PASSWORD) ?

5.5/mysql/bin/entrypoint.sh
```diff
@@ -0,0 +1,46 @@
...
+test -z "$MYSQL_DATABASE" && usage
+
+export X_SCLS="mysql55"
+source /opt/rh/mysql55/enable
+
+if [ ! -d '/var/lib/mysql/data' -a "$1" = 'mysqld_safe' ]; then
+
+	echo 'Running mysql_install_db ...'
+	mysql_install_db --user=mysql --datadir=/var/lib/mysql/data
+	echo 'Finished mysql_install_db'
```
mfojtik: I don't see the value of having 2 messages here other than you want to do a benchmark, in that case it is missing timestamps ;-)

Easiest would be to just print the one on line 22 and nuke the second

5.5/mysql/bin/entrypoint.sh
```diff
@@ -0,0 +1,46 @@
...
+
+	# These statements _must_ be on individual lines, and _must_ end with
+	# semicolons (no line breaks or comments are permitted).
+	# TODO proper SQL escaping on ALL the things D:
+	TEMP_FILE='/tmp/mysql-first-time.sql'
+	cat > "$TEMP_FILE" <<-EOSQL
+		DELETE FROM mysql.user ;
+		CREATE USER 'root'@'%' IDENTIFIED BY '${MYSQL_ROOT_PASSWORD}' ;
```
mfojtik: I see... are we going to give our users two mysql users? root && normal user? I would go just with one.

5.5/mysql/bin/entrypoint.sh
mnagy: It's so you can see where installation began and where it ended. mysql_install_db prints a lot of text; you want to be able to tell where it came from.

bparees:
this is still pending some investigation on mysqld_safe right?

mnagy:
This should be it. Tested for both RHEL and CentOS. The DB creation and option handling work. Entrypoint renamed to run-mysqld.sh, since that is the only thing it now does. You can pass options to it that get passed to mysqld.

mysqld also runs as pid 1 now. You can't pass SIGINT to it, it ignores it, since pid 1 doesn't have default signal handlers. But it listens to SIGTERM and will shutdown properly if received, so docker stop will work OK.

PTAL

mfojtik:
@mnagy looks good!

PR #2, opened 2015-02-24
Better trap logic for moving Dockerfiles around

mnagy:
Create a separate function that handles docker builds. This is only
required until Docker adds support for building with specific
Dockerfiles.

We setup a trap for ERR and RETURN signals. If the docker build fails,
ERR trap is triggered and Dockerfiles are restored. If it succeeds,
RETURN trap is triggered and the ERR trap is removed.

We'll want this in other repos as well.

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -8,12 +8,18 @@ declare -a VERSIONS=(5.5)
 OS=$1
 VERSION=$2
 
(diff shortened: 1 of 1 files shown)


bparees:
lgtm

mfojtik:
lgtm

PR #4, opened 2015-02-24
Separate MYSQL_USER from the root user

mnagy:
MySQL root user password can now be added optionally through the
MYSQL_ROOT_PASSWORD environment variable.

I know we talked about using 'admin', but it really seems to me that root is more common, e.g. in MySQL documentation.

I'm using the weird /dev/urandom magic because installing pwgen can be a real hassle (especially on rhel).

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -8,6 +8,7 @@ FROM centos:centos7
 #  * $MYSQL_USER - Database user name
 #  * $MYSQL_PASSWORD - User's password
 #  * $MYSQL_DATABASE - Name of the database to create
+#  * $MYSQL_ROOT_PASSWORD - MySQL root user password (optional)
 
 # Image metadata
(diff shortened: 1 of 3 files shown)


bparees:
hmm, so if they do not set it and instead we generate a random one that doesn't get recorded anywhere, are they still able to connect locally as the root user?

mfojtik:
I am ok with calling it root if that is what the documentation use.

LGTM

@bparees ?

mfojtik:
afaik you have to do this config when creating the user. (specify local no password)

bparees:
isn't that what we agreed to do?  if no root password env is set, configure root to be local only?

mfojtik:
yes it is, that was a hint to @mnagy how to add this behavior ;)

mnagy:
Will discuss this point with DB people today.

mnagy:
This PR is obsolete, please ignore. Please review instead PR #9.

PR #8, opened 2015-02-27
Add test script for the container

mnagy:


5.5/test/run
```diff
@@ -0,0 +1,69 @@
...
+		fi
+		attempt=$(( $attempt + 1 ))
+		sleep $sleep_time
+	done
+	echo "  Giving up: Failed to connect"
+	return 1
+}
+
+test_mysql() {
+	echo "  Testing table creation"
```
soltysh: You're actually doing create + insert + select + drop, so maybe some other comment than table creation?

soltysh:
Generally OK, but I think you're missing something like [hack/test.sh](https://github.com/openshift/sti-nodejs/blob/master/hack/test.sh) in this PR, because who is going to build the `openshift/mysql-55-centos7-candidate` for you?

mnagy:
Agreed. I think I'll try add the -candidate building logic into the Makefile. I don't want another shell script duplicating the functionality.

mnagy:
Added test target to Makefile. PTAL.
Note that I'd prefer not to squash this, but keep it as a separate commit.

soltysh:
LGTM.

soltysh:
LGTM :+1:

PR #9, opened 2015-03-03
Initialization sophistication

mnagy:
We now start mysqld after initialization, without network. We initialize
it, using CLI wherever possible instead of SQL files. After
initialization is done, we properly shutdown the server and then start
it again properly, using exec.
    
We also use 'source scl_source enable mysql55' instead of X_SCLS. It is
not supported as well, but should be a bit safer.

So far I've only tested this for CentOS. I have a problem with my RHEL setup atm and I want to enhance the test/run script before testing that.


mnagy:
[test]

mnagy:
Rebased on top of upstream/master. Please ignore jenkins test results, they work in PR #11 that enhances the tests and is based on this PR.

mnagy:
Closing, as per @mfojtik's offline request.

PR #11, opened 2015-03-03
Test images with and without root password

mnagy:
Also enhance the testing script, better cleanup.

Work is based on top of PR #9. Only last commit is original.

### DIFF
--- a/5.5/mysql/bin/run-mysqld.sh
+++ b/5.5/mysql/bin/run-mysqld.sh
@@ -1,43 +1,96 @@
 #!/bin/bash -e
 
+# Be paranoid and stricter than we should be.
+# https://dev.mysql.com/doc/refman/5.5/en/identifiers.html
+mysql_identifier_regex='^[a-zA-Z0-9_]+$'
+mysql_password_regex='^[a-zA-Z0-9_~!@#$%^&*()-=<>,.?;:|]+$'
+
 function usage {
 	echo "You must specify following environment variables:"
-	echo "  \$MYSQL_USER"
-	echo "  \$MYSQL_PASSWORD"
-	echo "  \$MYSQL_DATABASE"
+	echo "  \$MYSQL_USER (regex: '$mysql_identifier_regex')"
+	echo "  \$MYSQL_PASSWORD (regex: '$mysql_password_regex')"
+	echo "  \$MYSQL_DATABASE (regex: '$mysql_identifier_regex')"
+	echo "Optional:"
+	echo "  \$MYSQL_ROOT_PASSWORD (regex: '$mysql_password_regex')"
 	exit 1
 }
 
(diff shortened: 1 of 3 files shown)


5.5/mysql/bin/run-mysqld.sh
```diff
@@ -1,43 +1,96 @@
...
+	[[ "${var}" =~ $mysql_password_regex ]]
+}
+
+valid_mysql_identifier "$MYSQL_USER"     || usage
+valid_mysql_password   "$MYSQL_PASSWORD" || usage
+valid_mysql_identifier "$MYSQL_DATABASE" || usage
 
-chown -R mysql:mysql /var/lib/mysql
+# Make sure env variables don't propagate to mysqld process.
+mysql_user="$MYSQL_USER" ; unset MYSQL_USER
```
mfojtik: are you sure the env vars don't get propagated to mysql process this way? the lowercase vars will be the same....
i thought you are going to unset them when you starting the mysql server

5.5/mysql/bin/run-mysqld.sh
```diff
@@ -1,43 +1,96 @@
...
+
+	echo 'Running mysql_install_db'
+	mysql_install_db --user=mysql --datadir=/var/lib/mysql
+	# TODO: Not needed with --user=mysql. However, we should
+	#       strive to make this script runnable without root privileges
+	#chown -R mysql:mysql /var/lib/mysql
+
+	# Now start mysqld and add appropriate users.
+	echo 'Starting mysqld to create users'
+	/opt/rh/mysql55/root/usr/libexec/mysqld \
```
mfojtik: how about adding /opt/rh/mysql55/root/usr/libexec into PATH in Dockerfile?

mfojtik:
@mnagy good job! LGTM (see my 2 comments)

5.5/mysql/bin/run-mysqld.sh
mnagy: But these are not ENV vars (they are not exported). They are just ordinary bash variables, so they'll disappear on exec.
```
mnagy@ponik:~$ VAR=foo bash -c 'export' | grep VAR
declare -x VAR="foo"
mnagy@ponik:~$ VAR=foo bash -c 'VAR1=$VAR ; unset VAR ; bash -c export' | grep VAR
mnagy@ponik:~$
```

5.5/mysql/bin/run-mysqld.sh
mnagy: Just because of one or two execs? No.. It's libexec, it shouldn't be in PATH, IMO.

mnagy:
Squashed as per @mfojtik's request.

mfojtik:
LGTM

Issue #15, opened 2015-03-08
Security checkup


mnagy:
Make sure exactly which users get created during initialization, what their access rights are, and so on. This should be mainly done by looking into mysql.user table. We should also document as much as possible the behaviour of (not) setting `MYSQL_ROOT_USER`.

mnagy:
Relevant MySQL documentation:
* https://dev.mysql.com/doc/refman/5.5/en/adding-users.html
* https://dev.mysql.com/doc/refman/5.5/en/default-privileges.html

mnagy:
Reopening. The issue was only resolved partially, I'm going to add this to openshift-docs.

PR #16, opened 2015-03-08
Add more documentation into README.md

mnagy:


README.md
```diff
@@ -1,26 +1,83 @@
...
+Users can choose between RHEL and CentOS based images.
 
 # Installation and Usage
-
 Choose between CentOS7 or RHEL7 base image:
 
 *  **RHEL7 base image**
 
-To build a base-rhel7 image, you need to run docker build it on properly subscribed RHEL machine.
+To build a rhel7-based image, you need to run docker build on properly
```
bparees: on *a* properly subscribed...

README.md
```diff
@@ -1,26 +1,83 @@
...
+```
+
+If the database directory is not initialized, the entrypoint script will
+first run `mysql_install_db` and setup necessary database users and
+passwords. After the database is initialized, or if it was already
+present, `mysqld` is executed and will run as PID 1. You can stop the
+detached container by running `docker stop <CONTAINER ID>`.
+
+### MySQL root user
+The root user has no password set by default. You can set it by setting
```
bparees: you should describe the behavior they get if no root password is set (ie that you cannot remotely access the DB using that root account)..otherwise this is a very scary statement :)

bparees:
+1 to docs.

soltysh:
@bparees should this also land in openshift-docs? We have there [openshift_sti_image](https://github.com/openshift/openshift-docs/tree/master/openshift_sti_images) but I guess it also make sense to have this one as well and then rename that dir into openshift_images, what do you think?

bparees:
@soltysh yes it needs to go in openshift-docs, as for where, I think we should keep STI and non-STI separate, so maybe we need the top level section to be Images and the STI and Non-STI would be subsections.

soltysh:
+1

README.md
```diff
@@ -1,26 +1,83 @@
...
-This repository contains Dockerfiles for MySQL images for OpenShift. Users can choose between RHEL and CentOS based images.
-
+This repository contains Dockerfiles for MySQL images for OpenShift.
+Users can choose between RHEL and CentOS based images.
 
 # Installation and Usage
-
 Choose between CentOS7 or RHEL7 base image:
 
 *  **RHEL7 base image**
```
mfojtik: maybe we should change "base" to something else? "RHEL7 image"? The "base" sounds like a sti-base

README.md
```diff
@@ -1,26 +1,83 @@
...
+make build
 ```
-$ git clone https://github.com/openshift/mysql.git
-$ cd mysql
-$ make build
+
+## Environment variables and volumes
+
+The image recognizes following environment variables that you can set
+during initialization, by passing `-e VAR=VALUE` to the docker command.
```
mfojtik: Docker command?

README.md
```diff
@@ -1,26 +1,83 @@
...
+passwords. After the database is initialized, or if it was already
+present, `mysqld` is executed and will run as PID 1. You can stop the
+detached container by running `docker stop <CONTAINER ID>`.
+
+### MySQL root user
+The root user has no password set by default. You can set it by setting
+`MYSQL_ROOT_PASSWORD` environment variable when initializing your
+database.
+
+## Software collections
```
mfojtik: Software Collections

README.md
```diff
@@ -1,26 +1,83 @@
...
+We use [software collections](https://www.softwarecollections.org/) to
```
mfojtik: Same here

README.md
```diff
@@ -1,26 +1,83 @@
...
+install and launch MySQL. If you want to execute a command inside of a
+running container (for debugging for example), you need to prefix it
+with `scl enable` command. Some examples:
```
mfojtik: this sux, we should have the X_SCL approach in place here so you don't need to call "scl enable"

README.md
bparees: "based" instead of "base" probably fixes the problem, as seen on line 11.

README.md
bparees: "Docker `run` command"?

README.md
bparees: this is still important because otherwise i might try to run mysql commands directly (eg "docker run mycontainer mysql foo"), so they need to know that won't work.

README.md
mnagy: I know. As I was writing that, I realized I'm not really sure about the behaviour. I created issue #15 as a reminder for myself to check out what exactly happens and fix any possible problems and document them.

I guess we should create a card for this with a security label?

bparees: no you should probably resolve it as part of your mysql card.  it's a pretty fundamental functional issue.

mnagy:
Addressed suggestions. Docs got a PR as well.

bparees:
lgtm w/ the caveat that the no-root-password case still needs to be doc'd.

PR #18, opened 2015-03-17
Remove ugly {user,group}add

mnagy:
The Fedora/CentOS/RHEL mariadb/mysql rpms add mysql user and group in
the same manner we do (it was copied from it). It is not supposed to be
changed in future releases. Adding the user and group ourselves is very
ugly and the plethora of flags is confusing. Instead of that, we'll just
test that the UID is 27 in the Dockerfile. Even if UID in upstream image
would change, we'll get a build failure.

I realize this might be controversial (looking at you, @mfojtik), but I
really can't see any downside to be honest and people from the SCL team
that might end up maintaining this also were of the opinion that this
would be a better option.

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -18,13 +18,14 @@ ENV IMAGE_EXPOSE_SERVICES   3306:mysql
 MAINTAINER  Martin Nagy <mnagy@redhat.com>
 EXPOSE 3306
 
-RUN groupadd -g 27 -o -r mysql && \
-    useradd -M -N -g mysql -o -r -d /var/lib/mysql -s /sbin/nologin -c "MySQL Server" -u 27 mysql
-
+# This image must forever use UID 27 for mysql user so our volumes are
+# safe in the future. This should *never* change, the last test is there
+# to make sure of that.
 RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
     yum -y --setopt=tsflags=nodocs install https://www.softwarecollections.org/en/scls/rhscl/mysql55/epel-7-x86_64/download/rhscl-mysql55-epel-7-x86_64.noarch.rpm && \
     yum -y --setopt=tsflags=nodocs install hostname mysql55 && \
-    yum clean all
+    yum clean all && \
+    test "$(id mysql)" = "uid=27(mysql) gid=27(mysql) groups=27(mysql)"
 
 COPY run-mysqld.sh /usr/local/bin/
 COPY mysql /opt/openshift/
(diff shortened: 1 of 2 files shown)


bparees:
looks ok to me, but will wait for @mfojtik to weigh in before merging.

5.5/Dockerfile
```diff
@@ -18,14 +18,14 @@ ENV IMAGE_EXPOSE_SERVICES   3306:mysql
...
-RUN groupadd -g 27 -o -r mysql && \
-    useradd -M -N -g mysql -o -r -d /var/lib/mysql -s /sbin/nologin -c "MySQL Server" -u 27 mysql
-
 RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
     yum -y --setopt=tsflags=nodocs install https://www.softwarecollections.org/en/scls/rhscl/mysql55/epel-7-x86_64/download/rhscl-mysql55-epel-7-x86_64.noarch.rpm && \
     yum -y --setopt=tsflags=nodocs install hostname mysql55 && \
     yum clean all
 
+# Make sure the mysql user is installed with UID 27, so we're compatible across volumes.
+RUN test "$(id mysql)" = "uid=27(mysql) gid=27(mysql) groups=27(mysql)"
```
mfojtik: If you want to do this, you should do that in the previous step, not create a layer just for this check.
Again, I still think this is redundant and I think the previous approach was more clean and explicit (iow. WE create mysql user and WE control the UID of this user, so we can rely on this UID when we will do VOLUME for mysql data)

mfojtik:
@mnagy actually, user/group add is not ugly, it is explicit, it tells you what is created,etc...

5.5/Dockerfile
bparees: since the test is there, we're still controlling the uid.  and i tend to agree w/ the principal that we shouldn't be mucking about w/ things the RPM is trying to do.

I forget, what happens if we create the user but do NOT create it w/ uid 27, and then install the RPM?

bparees:
@mfojtik it's explicit, but ugly, imho. :)  (the args are pretty esoteric)

5.5/Dockerfile
mfojtik: @bparees this is not about RPM (AFAIK) but about having consistent and guaranteed UID on MySQL data volumes. Since we own the user, we can guarantee that all data created on MySQL volumes will use the same UID and this will not change in time.

bparees: @mfojtik yeah i understand your point there, but the reason @mnagy made this change, as I understand it, is because the RPM wants to create the user and requires the user be created w/ UID 27 (whether it creates it or you do).  

I prefer the approach of letting the RPM do it the way it wants rather than us trying to meet its requirements by creating the user manually.  (And then, as @mnagy has, a check so if the behavior ever changes, we catch it)

mnagy:
Forgot rhel7 dockerfile, DO NOT MERGE.

mnagy:
Should be ready for merge. PTAL

bparees:
lgtm.

PR #19, opened 2015-03-17
Move entrypoint to the context dir

mnagy:
/opt/openshift doesn't make much sense for just two files, move one of
them away, to /usr/local/bin. WORKDIR then becomes redundant. This also
serves as a workaround for issues we've seen on RHEL7 images ($PATH not
expanding in ENV instructions).

We also need to move my.cnf, I wanted to do that in a separate pull
request.


bparees:
lgtm.

jhadvig:
LGTM :+1:

PR #20, opened 2015-03-17
Fixing rhel7 bug with missing PATH Docker env var

jhadvig:
@mnagy @mfojtik PTAL


bparees:
is this needed with https://github.com/openshift/mysql/pull/19 going in?

jhadvig:
@bparees sorry figured out that no.

PR #21, opened 2015-03-18
Set password for root account to allow remote access.

### LINKED ISSUE #15: Security checkup

mnagy:
This partially resolves #15. What remains is update the openshift-docs.

Test work fine with both rhel and centos.

### DIFF
--- a/5.5/run-mysqld.sh
+++ b/5.5/run-mysqld.sh
@@ -85,10 +85,11 @@ if [ ! -d '/var/lib/mysql/mysql' ]; then
 	EOSQL
 
 	if [ -v root_pass ]; then
-		mysqladmin --socket=/tmp/mysql.sock password "${root_pass}" flush-privileges shutdown
-	else
-		mysqladmin --socket=/tmp/mysql.sock shutdown
+		mysql --socket=/tmp/mysql.sock <<-EOSQL
+			GRANT ALL PRIVILEGES ON *.* TO 'root'@'%' IDENTIFIED BY '${root_pass}';
+		EOSQL
 	fi
+	mysqladmin --socket=/tmp/mysql.sock flush-privileges shutdown
 fi
 
 exec /opt/rh/mysql55/root/usr/libexec/mysqld \
--- a/5.5/test/run
+++ b/5.5/test/run
@@ -83,6 +83,18 @@ create_container() {
 	echo "Created container $(cat $cidfile)"
 }
 
+assert_login_access() {
+	local USER=$1 ; shift
+	local PASS=$1 ; shift
+	local SUCCESS=$1 ; shift
+
+	if $SUCCESS; then
+		mysql_cmd <<< "SELECT 1;" && echo "    $USER($PASS) access granted as expected"
+	else
+		mysql_cmd <<< "SELECT 1;" || echo "    $USER($PASS) access denied as expected"
+	fi
+}
+
 run_tests() {
 	local name=$1 ; shift
 	envs="-e MYSQL_USER=$USER -e MYSQL_PASSWORD=$PASS -e MYSQL_DATABASE=db"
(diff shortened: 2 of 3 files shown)


bparees:
lgtm, please make sure the openshift-docs get this update too.

5.5/test/run
```diff
@@ -92,6 +104,17 @@ run_tests() {
...
 	test_connection $name
+	echo "  Testing login accesses"
+	assert_login_access $USER $PASS true
+	assert_login_access $USER "${PASS}_foo" false
+	if [ -v ROOT_PASS ]; then
+		assert_login_access root $ROOT_PASS true
+		assert_login_access root "${ROOT_PASS}_foo" false
+	else
+		assert_login_access root "foo" false
+		assert_login_access root "" false
```
bparees: hmm..  in the docs you imply that this should work: The root user has no password set by default, only allowing local
connections.

README.md
```diff
@@ -64,9 +64,11 @@ present, `mysqld` is executed and will run as PID 1. You can stop the
 detached container by running `docker stop <CONTAINER ID>`.
 
 ### MySQL root user
-The root user has no password set by default. You can set it by setting
-`MYSQL_ROOT_PASSWORD` environment variable when initializing your
-database.
+The root user has no password set by default, only allowing local
+connections. You can set it by setting `MYSQL_ROOT_PASSWORD` environment
```
jhadvig: Shouldn't it be the MYSQL_ADMIN_PASSWORD ?

5.5/run-mysqld.sh
```diff
@@ -85,10 +85,11 @@ if [ ! -d '/var/lib/mysql/mysql' ]; then
 	EOSQL
 
 	if [ -v root_pass ]; then
-		mysqladmin --socket=/tmp/mysql.sock password "${root_pass}" flush-privileges shutdown
-	else
-		mysqladmin --socket=/tmp/mysql.sock shutdown
+		mysql --socket=/tmp/mysql.sock <<-EOSQL
+			GRANT ALL PRIVILEGES ON *.* TO 'root'@'%' IDENTIFIED BY '${root_pass}';
```
mfojtik: Maybe do `FLUSH PRIVILEGES` here to assure the privileges are properly saved into data files?

README.md
mfojtik: +1 keep things consistent with Mongo and Postresql

mfojtik:
LGTM, two small nits

README.md
bparees: +1

mnagy: We've had a discussion about MySQL using the root user account, which is created by default and agreed by SCL people. Admin account wouldn't be the same as root, where root is more popular and also used by the official MySQL docs.

I don't see how consistency with mongo is relevant. Those are different worlds, most likely different users. And MYSQL_ROOT_PASSWORD is optional, mind you. If someone needs it, they should know what they are doing and they should read the docs where they'll discover that it's ROOT, not ADMIN.

5.5/test/run
mnagy: Those are not local connections, the mysql commands are executed from the host. By docker, using the same image, but from the host.

5.5/run-mysqld.sh
mnagy: Will fix.

5.5/test/run
bparees: got it.

README.md
bparees: /me is convinced.

mnagy:
Small change, added flush-privileges before shutdown.

README.md
jhadvig: Like to me it doesn't matter what would be the name, ROOT or ADMIN. The point is to stay consistent cause both of those roles, in mysql and mongo are doing the same. Why there should be a different names ?

mfojtik: let's issue this concern and return to it later ;-)

PR #22, opened 2015-03-19
Don't run the container as root and other improvements

mnagy:
This PR syncs MySQL image with some changes we introduced for PostgreSQL,
most notably it now allows the image to run under non-privileged user.
We now assume that the provided volume is writable. We also include a
writable directory in the image in case no other is provided.

Also some minor fixes and improvements to the test suite.

### DIFF
--- a/5.5/test/run
+++ b/5.5/test/run
@@ -19,13 +19,18 @@ cleanup() {
 
 		echo "Stopping and removing container $CONTAINER..."
 		docker stop $CONTAINER
+		exit_status=$(docker inspect -f '{{.State.ExitCode}}' $CONTAINER)
+		if [ "$exit_status" != "0" ]; then
+			echo "Dumping logs for $CONTAINER"
+			docker logs $CONTAINER
+		fi
 		docker rm $CONTAINER
 		rm $cidfile
 		echo "Done."
 	done
 	rmdir $CIDFILE_DIR
 }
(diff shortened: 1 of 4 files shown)


5.5/test/run
```diff
@@ -19,13 +19,18 @@ cleanup() {
 
 		echo "Stopping and removing container $CONTAINER..."
 		docker stop $CONTAINER
+		exit_status=$(docker inspect -f '{{.State.ExitCode}}' $CONTAINER)
+		if [ "$exit_status" != "0" ]; then
+			echo "Dumping logs for $CONTAINER"
+			docker logs $CONTAINER
+		fi
```
bparees: should actually exit w/ this exit_status right?  or some error code anyway.

mnagy: Hmm. Well, this is cleanup, so if it exited with non-zero status, the cleanup was probably called by trap and will exit with non-zero status anyway. I'm not sure what would happen if I'd put another exit there..

I think a situation where exit is normal but container exits with error shouldn't happen.

bparees: ok, that was my lack of knowledge about trap.  if trap is still going to exit w/ the code that caused the trap to be invoked, that's fine.  just wanted to make sure we weren't eating the error.

bparees:
lgtm.

PR #24, opened 2015-03-22
Various changes to sync-up with PostgreSQL image

mnagy:
This PR adds few improvements that were  novel in the PostgreSQL image.

* Put DB volume into /var/lib/mysql/data, add SCL .bashrc
* Move DB volume into /var/lib/mysql/data
* Add SCL enablement into /var/lib/mysql/.bashr so we can now run the
  arbitrary commands in the image without needing to override the
  entrypoint
* Test DB and user names for maximum allowed length
* Better error messages for invalid env vars
* Tests for the above

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -3,7 +3,7 @@ FROM centos:centos7
 # MySQL image for OpenShift.
 #
 # Volumes:
-#  * /var/lib/mysql - Datastore for MySQL
+#  * /var/lib/mysql/data - Datastore for MySQL
 # Environment:
 #  * $MYSQL_USER - Database user name
 #  * $MYSQL_PASSWORD - User's password
@@ -26,14 +26,16 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
     yum -y --setopt=tsflags=nodocs install https://www.softwarecollections.org/en/scls/rhscl/mysql55/epel-7-x86_64/download/rhscl-mysql55-epel-7-x86_64.noarch.rpm && \
     yum -y --setopt=tsflags=nodocs install hostname mysql55 && \
     yum clean all && \
-    mkdir -p /var/lib/mysql && chown mysql.mysql /var/lib/mysql && \
+    mkdir -p /var/lib/mysql/data && chown mysql.mysql /var/lib/mysql/data && \
     test "$(id mysql)" = "uid=27(mysql) gid=27(mysql) groups=27(mysql)"
 
 COPY run-mysqld.sh /usr/local/bin/
-COPY contrib /opt/openshift/
+COPY contrib/.bashrc /var/lib/mysql/
+COPY contrib/etc /opt/openshift/etc/
 
-VOLUME ["/var/lib/mysql"]
+VOLUME ["/var/lib/mysql/data"]
 
 USER mysql
 
 ENTRYPOINT ["run-mysqld.sh"]
+CMD ["mysqld"]
--- a/5.5/contrib/.bashrc
+++ b/5.5/contrib/.bashrc
@@ -1,0 +1,2 @@
+# This will make scl collection binaries work out of box.
+source scl_source enable mysql55
--- a/5.5/run-mysqld.sh
+++ b/5.5/run-mysqld.sh
@@ -1,11 +1,25 @@
-#!/bin/bash -e
+#!/bin/bash
+
+# For SCL enablement
+source $HOME/.bashrc
+
+set -eu
+
+# Data directory where MySQL database files live. The data subdirectory is here
+# because .bashrc lives in /var/lib/mysql/ and we don't want a volume to
+# override it.
+MYSQL_DATADIR=/var/lib/mysql/data
+MYSQL_DEFAULTS_FILE=/opt/openshift/etc/my.cnf
 
 # Be paranoid and stricter than we should be.
 # https://dev.mysql.com/doc/refman/5.5/en/identifiers.html
 mysql_identifier_regex='^[a-zA-Z0-9_]+$'
 mysql_password_regex='^[a-zA-Z0-9_~!@#$%^&*()-=<>,.?;:|]+$'
 
-function usage {
+function usage() {
+	if [ $# == 2 ]; then
+		echo "error: $1"
+	fi
 	echo "You must specify following environment variables:"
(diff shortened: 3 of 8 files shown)


5.5/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM centos:centos7
 # MySQL image for OpenShift.
 #
 # Volumes:
-#  * /var/lib/mysql - Datastore for MySQL
+#  * /var/lib/mysql/data - Datastore for MySQL
```
mfojtik: maybe explain the reasoning in docs or in comment

5.5/run-mysqld.sh
```diff
@@ -1,4 +1,9 @@
-#!/bin/bash -e
+#!/bin/bash
+
+# For SCL enablement
+. /var/lib/mysql/.bashrc
```
mfojtik: $HOME/.bashrc ? or is this script run as root?

maybe I would suggest to add a new variable 'MYSQL_DATA_PATH'

5.5/run-mysqld.sh
```diff
@@ -15,35 +23,24 @@ function usage {
...
-if [ "$MYSQL_ROOT_PASSWORD" ]; then
-	valid_mysql_password "$MYSQL_ROOT_PASSWORD" || usage
-	root_pass="$MYSQL_ROOT_PASSWORD"
-fi
-unset MYSQL_ROOT_PASSWORD
-
-# SCL in CentOS/RHEL 7 doesn't support --exec, we need to do it ourselves
-# The '|| exit 1' is here so -e doesn't propagate into scl_source.
-source scl_source enable mysql55 || exit 1
+function unset_env_vars {
```
mfojtik: can you add comment here as well? explain why we need to unset these env vars.

also can't you just write a one-liner? does unset support multiple env vars?

5.5/run-mysqld.sh
```diff
@@ -60,10 +57,13 @@ function wait_for_mysql {
...
 
-if [ ! -d '/var/lib/mysql/mysql' ]; then
+if [ "$1" = "mysqld" -a ! -d '/var/lib/mysql/data/mysql' ]; then
+
+	shift
+	check_env_vars
 
 	echo 'Running mysql_install_db'
-	mysql_install_db --datadir=/var/lib/mysql
+	mysql_install_db --datadir=/var/lib/mysql/data
```
mfojtik: i hate having hardcoded paths in scripts as they are harder to maintain... make "/var/lib/mysql" variable

5.5/run-mysqld.sh
```diff
@@ -78,21 +78,26 @@ if [ ! -d '/var/lib/mysql/mysql' ]; then
...
-			GRANT ALL PRIVILEGES ON *.* TO 'root'@'%' IDENTIFIED BY '${root_pass}';
+			GRANT ALL PRIVILEGES ON *.* TO 'root'@'%' IDENTIFIED BY '${MYSQL_ROOT_PASSWORD}';
 		EOSQL
 	fi
 	mysqladmin $admin_flags flush-privileges shutdown
+
+	unset_env_vars
+
+	exec /opt/rh/mysql55/root/usr/libexec/mysqld \
+		--defaults-file=/opt/openshift/etc/my.cnf \
```
mfojtik: make /opt/openshift env var

5.5/run-mysqld.sh
mnagy: It seems it does, will inline. Comment is already there :)

5.5/run-mysqld.sh
mnagy: I'll just add a defaults file env var, nothing else we use /opt/openshift for anyway, and it might go away soon in any case.

mnagy:
@mfojtik comments addressed. PTAL

soltysh:
LGTM.

5.5/contrib/.bashrc
```diff
@@ -0,0 +1,2 @@
+# This will make scl collection binaries work out of box.
+source scl_source enable mysql55
```
bparees: so does this work if i just docker exec into a running container with /bin/sh?

5.5/test/run
```diff
@@ -88,19 +88,58 @@ create_container() {
...
-	if $SUCCESS; then
+	if $success; then
 		mysql_cmd <<< "SELECT 1;" && echo "    $USER($PASS) access granted as expected"
 	else
 		mysql_cmd <<< "SELECT 1;" || echo "    $USER($PASS) access denied as expected"
 	fi
 }
 
-run_tests() {
+# Make sure the invocation of docker run fails.
```
bparees: image_creation seems like the wrong name for this... do you mean db creation?

5.5/test/run
```diff
@@ -88,19 +88,58 @@ create_container() {
...
+
+function try_image_invalid_combinations  {
+	assert_image_creation_fails "$@"
+	assert_image_creation_fails -e MYSQL_USER=user -e MYSQL_PASSWORD=pass "$@"
+	assert_image_creation_fails -e MYSQL_USER=user -e MYSQL_DATABASE=db "$@"
+	assert_image_creation_fails -e MYSQL_PASSWORD=pass -e MYSQL_DATABASE=db "$@"
+}
+
+function run_image_creation_tests {
+	echo "  Testing image entrypoint usage"
```
bparees: again this isn't image creation, it's container creation or db creation.

README.md
```diff
@@ -38,49 +34,23 @@ command.
...
+This will create a container named `mysql_database` running MySQL with database `db` and user with credentials `user:pass`. Port 3306 will be exposed and mapped to host. If you want your database to be persistent across container executions, also add a `-v /host/db/path:/var/lib/mysql/data` argument. This is going to be the MySQL data directory.
 
-### MySQL root user
-The root user has no password set by default, only allowing local
-connections. You can set it by setting `MYSQL_ROOT_PASSWORD` environment
-variable when initializing your container. This will allow you to login
-to the root account remotely. Local connections will still not require
-password.
-
-## Software Collections
```
bparees: it's probably still worth mentioning that the image is based on SCL.
1) it's a good plug
2) for people who docker exec into it, it may help them understand the directory structure better.
3) for people who docker exec commands directly (not running /bin/sh) they might need to know this.

5.5/contrib/.bashrc
mnagy: nope, but will work with interactive bash shell.

bparees:
lgtm.

PR #25, opened 2015-03-23
Generate configuration file with envsubst

mnagy:
This is my first take on https://trello.com/c/NGcr5u20/534-5-database-image-configuration-details-image-beta3

I added two settings from v2 MySQL cartridge. One, `MYSQL_MAX_CONNECTIONS` has a default set and is simply substituted in the config file. The other, `MYSQL_TIME_ZONE` is only used if it set, otherwise a commented entry is added into the config file for clarity.

Comments and suggestions are wanted, after this gets merged I'll do the same for PostgreSQL.


5.5/run-mysqld.sh
```diff
@@ -6,10 +6,14 @@ source $HOME/.bashrc
...
-# override it.
-MYSQL_DATADIR=/var/lib/mysql/data
-MYSQL_DEFAULTS_FILE=/opt/openshift/etc/my.cnf
+# because .bashrc and my.cnf both live in /var/lib/mysql/ and we don't want a
+# volume to override it.
+export MYSQL_DATADIR=/var/lib/mysql/data
+
+# Configuration settings.
+export MYSQL_DEFAULTS_FILE=$HOME/my.cnf
+export MYSQL_MAX_CONNECTIONS=${MYSQL_MAX_CONNECTIONS:-151}
```
bparees: rather than doing this, why not just default the values in the template file?

5.5/run-mysqld.sh
```diff
@@ -6,10 +6,14 @@ source $HOME/.bashrc
...
-# because .bashrc lives in /var/lib/mysql/ and we don't want a volume to
-# override it.
-MYSQL_DATADIR=/var/lib/mysql/data
-MYSQL_DEFAULTS_FILE=/opt/openshift/etc/my.cnf
+# because .bashrc and my.cnf both live in /var/lib/mysql/ and we don't want a
+# volume to override it.
+export MYSQL_DATADIR=/var/lib/mysql/data
+
+# Configuration settings.
+export MYSQL_DEFAULTS_FILE=$HOME/my.cnf
```
bparees: what's the intent here?  you later assume the file is really named my.cnf.template.

5.5/run-mysqld.sh
bparees: nm, i see envsubst doesn't support defaults.

5.5/run-mysqld.sh
```diff
@@ -94,6 +103,29 @@ if [ "$1" = "mysqld" -a ! -d "$MYSQL_DATADIR/mysql" ]; then
 		EOSQL
 	fi
 	mysqladmin $admin_flags flush-privileges shutdown
+}
+
+function generate_config() {
+	# Time zone.
```
bparees: i'd at least move your defaulting logic down to here so it's all in one place.

mnagy: I wanted all the logic at the top. Down in generate_config, the only defaulting logic is that if it's not set, it won't add any line in config (except for comment). There's a difference between setting having a default value and setting not being set at all.

5.5/run-mysqld.sh
mnagy: Yeah, the template has the .template suffix. But the generated file is `$MYSQL_DEFAULTS_FILE` and the variable is used multiple times as argument to different commands later.

README.md
```diff
@@ -34,6 +34,13 @@ The image recognizes following environment variables that you can set during ini
 |  `MYSQL_DATABASE`      | Database name                             |
 |  `MYSQL_ROOT_PASSWORD` | Password for the root user (optional)     |
 
+Following environment variables influence MySQL configuration file. They are all optional.
+
+|    Variable name         |    Description                                                  |    Default
+| :----------------------- | --------------------------------------------------------------- | -------------------------------
+|  `MYSQL_MAX_CONNECTIONS` | The maximum permitted number of simultaneous client connections |  151
+|  `MYSQL_TIME_ZONE`       | The MySQL server default timezone                               |  none
+
```
soltysh: I'd suggest reviewing [v2 mysql cart](https://github.com/openshift/origin-server/blob/master/cartridges/openshift-origin-cartridge-mysql/conf/my.cnf.erb) for variables that might be useful.

5.5/run-mysqld.sh
mfojtik: don't do this, check mysql docs for default TZ and set that, without this if/else. Also don't store configuration option in variables, only values...

5.5/contrib/my.cnf.template
```diff
@@ -0,0 +1,21 @@
...
+socket =
+
+# Disabling symbolic-links is recommended to prevent assorted security risks
+symbolic-links = 0
+
+# http://www.percona.com/blog/2008/05/31/dns-achilles-heel-mysql-installation/
+skip_name_resolve
+
+# Settings configured by the user:
+max_connections = ${MYSQL_MAX_CONNECTIONS}
```
mfojtik: ${MYSQL_MAX_CONNECTIONS:-123} doesn't work?

5.5/run-mysqld.sh
bparees: yeah, it's just not named well...  it's the mysql config file, right?  it's not the default or template.

mnagy: It is a defaults file (or option file) in MySQL nomenclature: https://dev.mysql.com/doc/refman/5.5/en/option-file-options.html#option_general_defaults-file

5.5/run-mysqld.sh
mnagy: There is no such thing. The default is that it's not in the config file at all. I could use another variable with another name, but that seems ugly and unnecessary (one more variable to keep track of..)

README.md
mnagy: Yes, I did that, but I'm not sure that's supposed to be part of this effort. I understood we only needed a system in place, we can add variables we'll want later.

5.5/contrib/my.cnf.template
mnagy: No.

README.md
bparees: yeah actually adding the config settings/env variables to the various DB carts should be part of this effort.

mnagy: Should I add all of them?

5.5/run-mysqld.sh
mfojtik: the fact it is not in config file does not mean it has no default ... check
mysql docs

mnagy: Default for optioned named "default-time-zone"? There is none. I know because I checked it not in documentation, but by running MySQL and asking with --help --verbose. Doc https://dev.mysql.com/doc/refman/5.5/en/server-options.html#option_mysqld_default-time-zone say that it's from system, usually from env var TZ.

README.md
bparees: that would be a good start... they all exist in the v2 cart because users asked for them over time, so it's likely the popular set.

5.5/contrib/my.cnf.template
bparees: envsubst doesn't handle defaulting unfortunately.  otherwise this whole thing could be much cleaner.

mnagy:
Added rest of options that we had in v2 and some tests. PTAL

5.5/run-mysqld.sh
bparees: huh, ok, looks good then! :)

bparees:
looking good @mnagy, i especially like the tests and doc!

mnagy:
Thanks! Docs where shamelessly stolen from v2 MySQL cartridge docs :)

soltysh:
LGTM

5.5/run-mysqld.sh
jhadvig: @bparees yeah, I was also a bit confused about the naming

5.5/contrib/my.cnf.template
```diff
@@ -0,0 +1,29 @@
...
+symbolic-links = 0
+
+# http://www.percona.com/blog/2008/05/31/dns-achilles-heel-mysql-installation/
+skip_name_resolve
+
+#
+# Settings configured by the user
+#
+
+${MYSQL_TIME_ZONE}
```
mfojtik: I still don't like this. We should use the env vars wisely to provide **values** to configuration directives... not to generate the directives themself. Can't you just add default-time-zone =${TZ} ? What happen when this setting will be just set to blank value? Is that default?

mfojtik:
NACK, I don't want us to use environment variables to contain the complete configuration directives as that is not clear and might be confusing. We need to stick to use them only as values for those directives. Even if that means that we will have to find some reasonable defaults for those ourself. Don't overuse envsubst to do templating.

soltysh:
I've must have missed that one taking back my LGTM, sorry. I agree with
@mfojtik that env var should be just for providing values and not entire
config lines.

mfojtik:
@soltysh @mnagy stick to customizations we can do easely with envsubts, everything else that require conditions or complex logic is out of scope. Users can provide their own my.cfg if they want just by downloading and customizing the image. So the simple rule is if you want to do something that require more logic than setting value for config directive, it is out-of-scope. @bparees agree?

mnagy:
Timezone dropped as agreed with @mfojtik. PTAL

mnagy:
Added comments in the template with default values as requested offline by @mfojtik.

README.md
```diff
@@ -34,6 +34,17 @@ The image recognizes following environment variables that you can set during ini
 |  `MYSQL_DATABASE`      | Database name                             |
 |  `MYSQL_ROOT_PASSWORD` | Password for the root user (optional)     |
 
+Following environment variables influence MySQL configuration file. They are all optional.
+
+|    Variable name                |    Description                                                    |    Default
+| :------------------------------ | ----------------------------------------------------------------- | -------------------------------
+|  `MYSQL_LOWER_CASE_TABLE_NAMES` | Sets how the table names are stored and compared                  |  0
+|  `MYSQL_DEFAULT_STORAGE_ENGINE` | The default storage engine (table type)                           |  InnoDB
```
mfojtik: this might break the VOLUME data if you update it for Pod. If the VOLUME already exists and it use ISAM and you change it to InnoDB the mysql won't start. Also in latest MySQL they are going to drop MyIsam so I think we should remove this option (dangerous and potentially going away in near future)

bparees:
We could still leave the timezone config line in the file, commented out..at least that would serve as a hint to users who wanted to configure it.

@mfojtik i'm not sure i agree..i'm ok with it of for now i guess but i'm not super happy about it, it's a regression from what we supported in v2.

mfojtik:
@bparees mysql allows to set timezone per connection or globally via SET TIMEZONE

bparees:
@mfojtik i'm not saying there aren't other ways to do it, but we did used to support it via config file too.

mnagy:
Removed storage engine setting.

mfojtik:
@bparees I agree, but I think we should also not blindly copy things from v2 just because we used to have them there. For example, users can set 'TZ' variable and mysql will pick it up during start. I'm just saying that initially, we don't need to have full-feature set for things we did support in v2, because the v2 was limited in what you can do in terms of customization.

mfojtik:
LGTM

soltysh:
LGTM

PR #30, opened 2015-04-15
Initial MySQL master->slave replication support

mfojtik:


mfojtik:
@danmcp PTAL

I need to make a README.md for this, but take a look on the JSON file provided if you are OK with this.
If you create this config, then you will end up with 1 Pod with 'mysql-master' and 3 Pod(s) with 'mysql-slave'.

If I connect to the 'mysql-master' Pod and run `SHOW SLAVE HOSTS;` command I get this:

```
mysql> show slave hosts;
+------------+-------------------+------+------------+
| Server_id  | Host              | Port | Master_id  |
+------------+-------------------+------+------------+
| 4294967295 | mysql-slave-4pqus | 3306 | 4098492312 |
|   16578557 | mysql-slave-b7eqs | 3306 | 4098492312 |
|  299383449 | mysql-slave-139bm | 3306 | 4098492312 |
+------------+-------------------+------+------------+
3 rows in set (0.00 sec)
```

mfojtik:
@mnagy PTAL as well

mfojtik:
FYI, the USER, PASSWORD and DATABASE settings are required only for 'mysqld-master' as the 'mysqld-slave' will replicate them. The only config both slave and master needs to share is the MASTER_USER and MASTER_PASSWORD so they can talk each-other.... I will provide documentation for all this ;-)

mfojtik:
@hhorak PTAL as well (I was talking about different entrypoints today, this is the PR)

5.5/contrib/common.sh
```diff
@@ -0,0 +1,155 @@
...
+mysql $mysql_flags <<-EOSQL
+    CREATE USER '${MYSQL_USER}'@'%' IDENTIFIED BY '${MYSQL_PASSWORD}';
+    GRANT ALL ON \`${MYSQL_DATABASE}\`.* TO '${MYSQL_USER}'@'%' ;
+    FLUSH PRIVILEGES ;
+EOSQL
+
+  if [ -v MYSQL_ROOT_PASSWORD ]; then
+mysql $mysql_flags <<-EOSQL
+      GRANT ALL PRIVILEGES ON *.* TO 'root'@'%' IDENTIFIED BY '${MYSQL_ROOT_PASSWORD}';
+EOSQL
```
mnagy: I think you can indent the mysql command itself, but the GRANT line should have no spaces in the beginning. Also, you can remove the - from <<-EOF since that causes preceding tabs (but not spaces) to be ignored.

mfojtik:
@danmcp @bparees tests passed, this is ready for merge now.

5.5/Dockerfile
```diff
@@ -24,12 +24,15 @@ EXPOSE 3306
...
     yum -y --setopt=tsflags=nodocs install https://www.softwarecollections.org/en/scls/rhscl/mysql55/epel-7-x86_64/download/rhscl-mysql55-epel-7-x86_64.noarch.rpm && \
-    yum -y --setopt=tsflags=nodocs install gettext hostname mysql55 && \
+    yum -y --setopt=tsflags=nodocs install gettext hostname bind-utils mysql55 && \
     yum clean all && \
     mkdir -p /var/lib/mysql/data && chown mysql.mysql /var/lib/mysql/data && \
     test "$(id mysql)" = "uid=27(mysql) gid=27(mysql) groups=27(mysql)"
 
 COPY run-mysqld.sh /usr/local/bin/
+COPY run-mysqld-master.sh /usr/local/bin/
+COPY run-mysqld-slave.sh /usr/local/bin/
```
bparees: no way to do this in one layer by pulling a whole directory in?

5.5/examples/replica/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+If you want to perform administration tasks, you can also set the
+`MYSQL_ROOT_PASSWORD`. In that case you will be able to connect to the MySQL
+server as the 'root' user and create more users or more databases.
+
+Once the MySQL master server is started, it has no slaves preconfigured as the
+slaves registers automatically.
+
+Note that currently the multiple-master configuration is not supported (even
+though the `mysql-master` is defined as ReplicationController. If you bump the
```
bparees: s/bump/increase/

5.5/Dockerfile
mfojtik: @bparees I will have to add these into sub-directory, which is not a bad idea actually as the upstream/atomic is moving forward to standardize 'entrypoint.d' or something similar.

5.5/examples/replica/teardown.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+osc update rc mysql-master --patch='{ "apiVersion": "v1beta1", "desiredState": { "replicas": 0 }}'
```
bparees: is there any benefit to setting RC to 0 before destroying it anyway?

5.5/Dockerfile
mfojtik: @bparees I fixed it, just COPY run-*.sh /usr/local/bin so we are back in one layer.

mnagy:
LGTM

5.5/examples/replica/teardown.sh
mfojtik: @bparees without 0 the Pods will stay running (as far as I can see)

bparees: yeah i was thinking you were deleting the pods explicitly, nm.

bparees:
lgtm.

PR #32, opened 2015-04-22
Support random uid at runtime

csrwng:
Allows running this container with any uid by making the home/data directory
world-writeable and setting the HOME directory explicitly in the startup script. 
Example:
```
docker run -p 3306:3306 -u 12345 -e MYSQL_USER=user -e MYSQL_PASSWORD=pass \
   -e MYSQL_DATABASE=db1 -e HOME=/var/lib/mysql openshift/mysql-55-centos7
```

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -33,9 +33,13 @@ COPY run-*.sh /usr/local/bin/
 
 COPY contrib /var/lib/mysql/
 
+RUN chmod o+w -R /var/lib/mysql
+
 VOLUME ["/var/lib/mysql/data"]
 
 USER mysql
+
+ENV HOME /var/lib/mysql
 
 ENTRYPOINT ["run-mysqld.sh"]
 CMD ["mysqld"]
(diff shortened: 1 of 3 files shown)


csrwng:
@mnagy @jhadvig please review

bparees:
[test]

bparees:
add to whitelist

5.5/contrib/common.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# Explicitly set the home directory to allow arbitrary uids to use this image
+export HOME=/var/lib/mysql
```
bparees: hmm...what will this mean for docker exec type operations?  they'll have no homedir, which also means the .bashrc won't be processed, which means they won't have the SCL stuff enabled either.

short of creating the passwd file, i'm not sure there's anything else we can do, but.. something to think about.

mnagy: setting $HOME in the Dockerfile?

csrwng: Setting HOME in the Dockerfile is a good idea. I've updated the PR to do that.

csrwng:
Added random uid tests

mnagy:
LGTM

jhadvig:
LGTM

PR #46, opened 2015-06-09
Make mysql command work out of the box

### LINKED ISSUE #45: Make mysql command work out of box
Currently, running mysql command in a running container requires a lot of command line arguments, like hostname, username and password. I would be in favour of enabling the unix socket. This is the easiest way to do it and is the same way we do it with PostgreSQL.

Also, I'd like to create a bash script in standard path named "mysql" that will scl_source the collection and do an `exec mysql`. This way `docker exec` -ti mysql as well as `oc exec` will work out of box and will in my opinion greatly enhance the user experience. This I think we should also do for PostgreSQL and MongoDB images.

@mfojtik @bparees thoughts? cc @hhorak

mnagy:
Enable mysql unix socket and set BASH_ENV so that docker
`exec -ti CONTAINER_ID bash -c mysql` works correctly.

Fixes #45

@mfojtik @hhorak

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -13,7 +13,8 @@ FROM centos:centos7
 MAINTAINER  Martin Nagy <mnagy@redhat.com>
 
 ENV MYSQL_VERSION=5.5 \
-    HOME=/var/lib/mysql
+    HOME=/var/lib/mysql \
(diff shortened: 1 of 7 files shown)


mfojtik:
@mnagy LGTM but wait for the test... we have to do same for other images

PR #48, opened 2015-06-11
Better SCL enablement, tests

mnagy:
Details at https://github.com/openshift/sti-base/issues/38

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -13,8 +13,7 @@ FROM centos:centos7
 MAINTAINER  Martin Nagy <mnagy@redhat.com>
 
 ENV MYSQL_VERSION=5.5 \
-    HOME=/var/lib/mysql \
-    BASH_ENV=/var/lib/mysql/.bashrc
+    HOME=/var/lib/mysql
 
 LABEL k8s.io/description="MySQL is a multi-user, multi-threaded SQL database server" \
       k8s.io/display-name="MySQL 5.5" \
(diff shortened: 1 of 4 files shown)


5.5/Dockerfile
```diff
@@ -37,6 +36,13 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
     test "$(id mysql)" = "uid=27(mysql) gid=27(mysql) groups=27(mysql)" && \
     chmod o+w -R /var/lib/mysql
 
+# When bash is started non-interactively, to run a shell script, for example it
+# looks for this variable and source the content of this file. This will enable
+# the SCL for all scripts without need to do 'scl enable'.
+ENV BASH_ENV=/var/lib/mysql/scl_enable \
+    ENV=/var/lib/mysql/scl_enable \
+    PROMPT_COMMAND=". /var/lib/mysql/scl_enable"
+
```
bparees: I guess this is ok because we're squashing but any reason you didn't include this in the env section above on line 16?

5.5/Dockerfile.rhel7
```diff
@@ -38,6 +37,14 @@ RUN yum install -y yum-utils gettext hostname && \
     test "$(id mysql)" = "uid=27(mysql) gid=27(mysql) groups=27(mysql)" && \
     chmod o+w -R /var/lib/mysql
 
+# When bash is started non-interactively, to run a shell script, for example it
+# looks for this variable and source the content of this file. This will enable
+# the SCL for all scripts without need to do 'scl enable'.
+ENV BASH_ENV=/var/lib/mysql/scl_enable \
+    ENV=/var/lib/mysql/scl_enable \
+    PROMPT_COMMAND=". /var/lib/mysql/scl_enable"
```
bparees: same.

5.5/Dockerfile
mnagy: Mostly because it's logically separate functionality than setting MYSQL_VERSION and HOME, and looks better with the three line comment.

mfojtik: @bparees I want to have comment on this to explain why we are doing it... squashing should take care of this

PR #52, opened 2015-06-18
Fix the mysql replication example

mfojtik:


mfojtik:
@bparees this just updates the template to adopt recent patterns. I tested this and it is working \o/

mfojtik:
@jhadvig PTAAH

jhadvig:
@mfojtik looking && testing...

bparees:
@jhadvig testing successful?

mfojtik:
I tested this and presented today :) Everything is working and you can
nicely scale this up or down... One thing that I think I should change is
the service name to be just mysql so you can use this template with our
examples.

bparees:
you can use the template w/ our examples, our examples let you specify the DB service name now.

PR #55, opened 2015-06-19
Remove trap in ./hack/build.sh

mfojtik:
https://github.com/openshift/sti-base/issues/43

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -16,24 +16,6 @@ function squash {
   $HOME/.local/bin/docker-scripts squash -f $base ${IMAGE_NAME}
 }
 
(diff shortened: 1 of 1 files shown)


bparees:
if you're going to do this, let's rename them to Dockerfile.centos7 too.

bparees:
though to do that you're going to have to fix up vagrant too.

mfojtik:
@bparees would prefer to have at least one 'Dockerfile' in repo ;-) So if you just do docker build it works without having to specify extra parameters or use `make build`.

bparees:
@mfojtik i guess.  could symlink dockerfile->dockerfile.centos7 :)

PR #60, opened 2015-06-30
Fix labels to match the new schema

mfojtik:


bparees:
again why only preserve in rhel?

PR #63, opened 2015-07-08
Some fixes to MySQL replication example

mfojtik:
This PR will add `PersistentVolume` support to the MySQL replication example for the MySQL master server. 
This also allows users to change the password for the MySQL master by using `oc edit dc/mysql-master`.

### DIFF
--- a/5.5/contrib/common.sh
+++ b/5.5/contrib/common.sh
@@ -70,19 +70,21 @@ function wait_for_mysql() {
   done
 }
 
+function start_local_mysql() {
+  # Now start mysqld and add appropriate users.
+  echo 'Starting local mysqld server ...'
+  /opt/rh/mysql55/root/usr/libexec/mysqld \
+    --defaults-file=$MYSQL_DEFAULTS_FILE \
+    --skip-networking --socket=/tmp/mysql.sock &
+  mysql_pid=$!
+  wait_for_mysql $mysql_pid
+}
+
 # Initialize the MySQL database (create user accounts and the initial database)
 function initialize_database() {
   echo 'Running mysql_install_db ...'
   mysql_install_db --datadir=$MYSQL_DATADIR
-
-  # Now start mysqld and add appropriate users.
-  echo 'Starting mysqld to create users ...'
-  /opt/rh/mysql55/root/usr/libexec/mysqld \
-    --defaults-file=$MYSQL_DEFAULTS_FILE \
-    --skip-networking --socket=/tmp/mysql.sock &
-
-  mysql_pid=$!
-  wait_for_mysql $mysql_pid
+  start_local_mysql
 
   [ -v MYSQL_DISABLE_CREATE_DB ] && return
 
--- a/5.5/examples/replica/README.md
+++ b/5.5/examples/replica/README.md
@@ -19,8 +19,15 @@ See: https://dev.mysql.com/doc/refman/5.5/en/replication.html
 
 ## How does this example work?
 
-The provided JSON file (`mysql_replica.json`) contains a `Config` resource that
+The provided JSON file (`mysql_replica.json`) contains a `Template` resource that
 groups the Kubernetes and OpenShift resources which are meant to be created.
+This template will start with one MySQL master server and one slave server.
+
+## Persistent storage
+
+In order to provide the persistent storage for the MySQL master server, the administrator
+of OpenShift needs to create a PersistentVolume that you can claim. This example requires a PersistentVolume of size 512m be available.
+To learn more about how to create PersistentVolume, refer to [OpenShift documentation](https://docs.openshift.org/latest/admin_guide/persistent_storage_nfs.html)
 
 ### Service 'mysql-master'
 
--- a/5.5/examples/replica/mysql_replica.json
+++ b/5.5/examples/replica/mysql_replica.json
@@ -47,6 +47,23 @@
   ],
   "objects": [
     {
+      "kind": "PersistentVolumeClaim",
+      "apiVersion": "v1",
+      "metadata": {
+        "name": "mysql-master"
+      },
+      "spec": {
+        "accessModes": [
+          "ReadWriteOnce"
+        ],
+        "resources": {
+          "requests": {
+            "storage": "512Mi"
+          }
+        }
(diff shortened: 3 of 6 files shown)


mfojtik:
@bparees @mnagy PTAL

5.5/contrib/common.sh
```diff
@@ -70,20 +70,22 @@ function wait_for_mysql() {
...
 
-# Initialize the MySQL database (create user accounts and the initial database)
-function initialize_database() {
-  echo 'Running mysql_install_db ...'
-  mysql_install_db --datadir=$MYSQL_DATADIR
-
+function start_local_mysql() {
   # Now start mysqld and add appropriate users.
-  echo 'Starting mysqld to create users ...'
+  echo 'Starting local mysqld to  ...'
```
mnagy: to what?

5.5/examples/replica/mysql_replica.json
```diff
@@ -47,6 +47,23 @@
   ],
   "objects": [
     {
+      "kind": "PersistentVolumeClaim",
+      "apiVersion": "v1beta3",
```
mnagy: Shouldn't it be v1?

5.5/examples/replica/persistent_nfs_volume.json
```diff
@@ -0,0 +1,20 @@
+{
```
bparees: rather than provide this (because do we really want to provide this in every sample that needs a persistent volume?) i'd prefer to see us link to good documentation about how to create persistent volumes.

bparees: not to mention it crosses the boundary between what a user trying to run this sample needs to do, and what an admin setting up an openshift environment needs to do.

5.5/test/run
```diff
@@ -89,6 +89,39 @@ function create_container() {
   echo "Created container $(cat $cidfile)"
 }
 
+function run_replication_test() {
```
bparees: can we get a test case that 
1) launches the image using a local volume (docker run, not in openshift)
2) shuts it down
3) launches it using a new password
4) confirms the new password now works and the old one no longer does?

5.5/examples/replica/persistent_nfs_volume.json
mfojtik: @bparees that boundary is very thin here as this is intended for users to try the replication template. That template won't work until the users (or admins) create the volume....

5.5/test/run
mfojtik: yes, but I will add it as a separate test case (not part of replication)

bparees: yeah i was just lazy about where i put the comment :)

5.5/examples/replica/persistent_nfs_volume.json
bparees: i understand, but i still don't think the sample should be responsible for that.  persistent volumes are a general openshift concept, not something specific to this sample, so i don't want to duplicate content related to them.

mfojtik: updated docs :)

On Wed, Jul 8, 2015 at 3:45 PM, Ben Parees <notifications@github.com> wrote:

> In 5.5/examples/replica/persistent_nfs_volume.json
> <https://github.com/openshift/mysql/pull/63#discussion_r34148425>:
>
> > @@ -0,0 +1,20 @@
> > +{
>
> i understand, but i still don't think the sample should be responsible for
> that. persistent volumes are a general openshift concept, not something
> specific to this sample, so i don't want to duplicate content related to
> them.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openshift/mysql/pull/63/files#r34148425>.
>



-- 
-----------------------------------------------
Michal Fojtik <mi@mifo.sk <mfojtik@mifo.sk>>
Red Hat OpenShift, Engineering

5.5/examples/replica/README.md
```diff
@@ -19,8 +19,15 @@ See: https://dev.mysql.com/doc/refman/5.5/en/replication.html
...
-The provided JSON file (`mysql_replica.json`) contains a `Config` resource that
+The provided JSON file (`mysql_replica.json`) contains a `Template` resource that
 groups the Kubernetes and OpenShift resources which are meant to be created.
+This template will start with one MySQL master server and one slave server.
+
+## Persistent storage
+
+In order to provide the persistent storage for the MySQL master server, administrator
+of OpenShift needs to create a PersistentVolume that you can claim. To learn more
+about how to create PersistentVolume, refer to [OpenShift documentation](https://docs.openshift.org/latest/admin_guide/persistent_storage_nfs.html)
```
bparees: s/administrator/the administrator/

and add a sentence about "this example requires a persistent volume of size 512m be available"

mfojtik:
@bparees thanks! all comments addressed... have to wait for tests to pass.

5.5/test/run
```diff
@@ -89,6 +89,55 @@ function create_container() {
...
+  # Create MySQL container with persistent volume
+  create_container "testpass1" -e MYSQL_USER=user -e MYSQL_PASSWORD=foo \
+    -e MYSQL_DATABASE=db -v ${tmpdir}:/var/lib/mysql/data:Z
+  CONTAINER_IP=$(get_container_ip testpass1) USER=user PASS=foo test_connection "testpass1"
+
+  docker stop $(get_cid testpass1)
+
+  create_container "testpass2" -e MYSQL_USER=user -e MYSQL_PASSWORD=bar \
+    -e MYSQL_DATABASE=db -v ${tmpdir}:/var/lib/mysql/data:Z
+  CONTAINER_IP=$(get_container_ip testpass2) USER=user PASS=bar test_connection "testpass2"
```
bparees: you should negative test that "testpass1" doesn't work anymore, for completion.

mfojtik: that container is stopped... otherwise two mysql servers will share same folder :)

mfojtik: nvd I see your point

mfojtik:
@bparees merge?

bparees:
lgtm.

PR #64, opened 2015-07-10
Mysql 5.6 with GTID based replication

hhorak:
This is initial commit to add mysql 5.6 based image. One commit just changes the packages the image is based on, another commit changes replication implementation to use GTID, which allows easier replication management (no binlog files/locations needed).

### DIFF
--- a/5.6/Dockerfile
+++ b/5.6/Dockerfile
@@ -1,0 +1,51 @@
+FROM centos:centos7
+
+# MySQL image for OpenShift.
+#
+# Volumes:
+#  * /var/lib/mysql/data - Datastore for MySQL
+# Environment:
+#  * $MYSQL_USER - Database user name
+#  * $MYSQL_PASSWORD - User's password
+#  * $MYSQL_DATABASE - Name of the database to create
+#  * $MYSQL_ROOT_PASSWORD (Optional) - Password for the 'root' MySQL account
+
+MAINTAINER  Martin Nagy <mnagy@redhat.com>
+
+ENV MYSQL_VERSION=5.6 \
+    HOME=/var/lib/mysql
+
+LABEL io.k8s.description="MySQL is a multi-user, multi-threaded SQL database server" \
+      io.k8s.display-name="MySQL 5.6" \
+      io.openshift.expose-services="3306:mysql" \
+      io.openshift.tags="database,mysql,mysql56,rh-mysql56"
+
+EXPOSE 3306
+
+COPY run-*.sh /usr/local/bin/
+COPY contrib /var/lib/mysql/
+
+# This image must forever use UID 27 for mysql user so our volumes are
+# safe in the future. This should *never* change, the last test is there
+# to make sure of that.
+RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
+    yum -y --setopt=tsflags=nodocs install https://www.softwarecollections.org/en/scls/rhscl/rh-mysql56/epel-7-x86_64/download/rhscl-rh-mysql56-epel-7-x86_64.noarch.rpm && \
+    yum -y --setopt=tsflags=nodocs install gettext hostname bind-utils rh-mysql56 && \
+    yum clean all && \
+    mkdir -p /var/lib/mysql/data && chown mysql.mysql /var/lib/mysql/data && \
+    test "$(id mysql)" = "uid=27(mysql) gid=27(mysql) groups=27(mysql)" && \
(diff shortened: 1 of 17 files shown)


hhorak:
Almost forgot -- `docker logs` calls in the test were just meant to make debugging easier, ideally should be in separate patch, but it was late, so it sneaked in somehow :)

mfojtik:
test

5.5/test/run
```diff
@@ -140,6 +140,7 @@ function run_replication_test() {
       echo "The ${slave_ip} failed to register in MASTER"
       return 1
     fi
+    docker logs $(get_cid slave.cid)
```
bparees: is this critical as status/progress information, or can you make this only dump out when there's a failure?  we already dump container logs in cleanup() on failures, but not the slave container currently of course.

bparees:
with this PR you can update the https://github.com/openshift/mysql/blob/master/README.md to indicate we support more than 5.5 :)

5.5/test/run
hhorak: Good point, I'll change it so the logs are dumped only on failure.

mfojtik:
test

5.6/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+MAINTAINER  Martin Nagy <mnagy@redhat.com>
+
+ENV MYSQL_VERSION=5.6 \
+    HOME=/var/lib/mysql
+
+LABEL io.k8s.description="MySQL is a multi-user, multi-threaded SQL database server" \
+      io.k8s.display-name="MySQL 5.6" \
+      io.openshift.expose-services="3306:mysql" \
+      io.openshift.tags="database,mysql,rh-mysql56"
```
mfojtik: why is this rh-mysql56?

5.6/Dockerfile.rhel7
```diff
@@ -0,0 +1,64 @@
...
+
+MAINTAINER  Martin Nagy <mnagy@redhat.com>
+
+ENV MYSQL_VERSION=5.6 \
+    HOME=/var/lib/mysql
+
+LABEL k8s.io/description="MySQL is a multi-user, multi-threaded SQL database server" \
+      k8s.io/display-name="MySQL 5.6" \
+      openshift.io/expose-services="3306:mysql" \
+      openshift.io/tags="database,mysql,rh-mysql56"
```
mfojtik: same here

mfojtik:
test

mfojtik:
@hhorak I get this error when running the image: https://gist.github.com/mfojtik/c1bf1bc2b79c66a8d2b7

Also the test will fail.

mnagy:
@mfojtik iirc CTRL+C never worked for MySQL (no signal handler).

mfojtik:
@mnagy then it is a bug

mnagy:
@mfojtik it's not our fault. PID 1 will not have default signal handlers installed and MySQL does not install SIGINT handler by itself. docker stop works however, that's what is truly important for OS.

mfojtik:
@mnagy being able to SIGINT from docker run is also important for user experience... (not that important than running in OS ;-) Would be nice to have trap or something that send signal to mysql. Not a high prio.

mnagy:
@mfojtik not possible. Trap would only apply for the bash process, upon exec the signal mask is disabled. Trust me, I've researched this when first doing the MySQL image. We'd need to patch MySQL itself.

mfojtik:
@hhorak can you please fix the annotations to match with mysql55 image and squash the commits? 

LGTM and jenkins is also happy about this PR.

@mnagy will run second round review.

5.6/Dockerfile
hhorak: rh-mysql56 is name of the collection the image is based on, so let's use both, mysql56 and rh-mysql56 so the search works fine for both cases.

hhorak:
I had some troubles with squashing, but it seems to be ready.

mfojtik:
test

hhorak:
@mfojtik, as for killing mysqld using `SIGINT`, @mnagy is right, it is actually a feature that mysqld is not possible to kill using this signal. We may work with upstream to add an option to disable it, that would be used in containers. Do you have some particular use case where you'd need `SIGINT` working or is it just that it works with other images?

mnagy:
LGTM, but please do not merge until I test the replication with OS.

mnagy:
Verified this works with OS. I've noticed the 5.5 replication didn't work for me, for some reason :/

Merging.

PR #66, opened 2015-07-13
Update docker scripts

mfojtik:


mfojtik:
@bparees to give a context here... the `docker_py-1.3.0` was released and since that the squashing tool is broken. I reported the issue https://github.com/goldmann/docker-scripts/issues/35 but I think it would be more safe to hardcode the version that is known to work to avoid future breakage.

PR #69, opened 2015-07-22
Refactoring scripts and making the image easily extensible

hhorak:
This is mostly a redesign of the scripts, while most changes were already consulted before, but never made it to these docker images so far.

Most of the changes are transparent for the image users, while they are just re-organizing the scripts. The reason for the changes is to allow to extend images easier for user and also to make future maintenance simpler, since one complex bash script would be hard to change in another layer.

These are the changes made:
* available commands (that are expected to be used as docker run argument) are now called `run-mysqld`, `run-mysqld-master`, `run-mysqld-slave` and those binaries are available under `/usr/bin`, so users may start them on demand, e.g. when debugging the image (if `run-mysqld-master.sh` or `mysqld-master` are already considered to be API, we can add those symlinks to not break depended stuff)
* entrypoint script is now called `container-entrypoint`
* the entrypoint now serves just for changing environment, nothing more
* configuration files are located in the path that binaries expect (e.g. `/opt/rh/mysql55/root/etc/my.cnf`) and `/etc/my.cnf`, `/etc/my.cnf.d` are symlinks to those files, so users may work with `/etc/my.cnf` and `/etc/my.cnf.d` directly. As a result, `--defaults-file` doesn't have to be used when working with binaries
* preparation of directories during build (fixing access) is now done by script `/usr/libexec/cont-setup`, so dockerfile is more clean
* in configuration file, the `mysql` user is not specified, we want to let running daemon under actual user every-time
* images are now more easily extensible, by dropping scripts with further initialization into expected directories (e.g. `/usr/share/cont-layer/mysql/post-init`)
* scripts don't include collection name hard-coded, but the collection is specified in the Dockerfile and thus most of the scripts may be common for more versions of mysql (which makes maintenance easier)
* `container-usage` returns help for the container


mfojtik:
test

mfojtik:
@mnagy can you review this?

5.5/root/usr/libexec/cont-setup
```diff
@@ -0,0 +1,58 @@
...
+    ln -s ${MYSQL_CONFIG_FILE} /etc/my.cnf
+    ln -s ${MYSQL_CONFIG_FILE}.d /etc/my.cnf.d
+fi
+
+# setup directory for data
+for dir in /var/lib/mysql{,/data} ; do
+    mkdir -p $dir
+    chown -R mysql:mysql $dir
+    restorecon -R $dir
+done
```
mnagy: No need for the loop here, just mkdir -p /var/lib/mysql/data and chown&restorecon -R /var/lib/mysq.

5.5/Dockerfile
```diff
@@ -32,20 +29,23 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
     yum -y --setopt=tsflags=nodocs install https://www.softwarecollections.org/en/scls/rhscl/mysql55/epel-7-x86_64/download/rhscl-mysql55-epel-7-x86_64.noarch.rpm && \
     yum -y --setopt=tsflags=nodocs install gettext hostname bind-utils mysql55 && \
     yum clean all && \
-    mkdir -p /var/lib/mysql/data && chown mysql.mysql /var/lib/mysql/data && \
```
mnagy: What happens if the image is run with a random uid and /var/lib/mysql/data is not mounted (and hence, not accessible?)

mnagy:
A note about duplication: The 5.5 and 5.6 share almost everything and it's all duplicated. Can we do something about that? Related to that, I think it's a bit weird that we have MYSQL_SCL env var, since it doesn't contribute to duplication avoidance.

5.5/root/usr/libexec/cont-setup
hhorak: yeah, there were more of them at one point, but after I remove them I didn't remove the for.. I will fix it..

5.5/Dockerfile
hhorak: It will be accessible even if not mounted, it is created in the `/usr/libexec/cont-setup`, that is run during the docker build. And `chmod o+w -R /var/lib/mysql` in the same file should handle the access for random uid. Maybe I don't see which use case you see to be an issue?

mnagy: I didn't see you were running cont-setup during Docker build, thanks for pointing that out.

5.5/root/usr/share/cont-lib/cont-lib.sh
```diff
@@ -0,0 +1,99 @@
...
+cont_debug2() { __cont_dbg 2 "$*" ; }
+cont_debug3() { __cont_dbg 3 "$*" ; }
+
+
+__cont_encode_env()
+{
+    local i
+    for i in $1
+    do
+        eval local val="\$$i"
```
mnagy: Is the eval necessary here?

mnagy:
What is the purpose of `parser-simple-config.sh`? It's extremely scary, writing a config file parser in bash doesn't seem like a good idea, and besides, it doesn't seem to be used anywhere.

5.5/root/usr/share/cont-lib/cont-lib.sh
praiskup: Yes, $val should be set to value of variable named $i.  If `$i = VARNAME` then `val=$VARNAME`.

mnagy:
One more note about duplication: It seems that lot of these library scripts should be put into a separate common Docker image.

5.5/Dockerfile
hhorak: RUN `/usr/libexec/cont-setup` -- line 44

hhorak:
`parser-simple-config.sh` was used before to be able to have more options in the configuration changed. It is now not used, so I'm fine with removing this script for now.

5.5/Dockerfile
```diff
@@ -32,20 +29,23 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
...
-    ENV=/var/lib/mysql/scl_enable \
-    PROMPT_COMMAND=". /var/lib/mysql/scl_enable"
+ENV BASH_ENV=/usr/share/cont-lib/cont-env.sh \
+    ENV=/usr/share/cont-lib/cont-env.sh \
+    PROMPT_COMMAND=". /usr/share/cont-lib/cont-env.sh" \
+    MYSQL_SCL=mysql55
+
+ADD root /
+
+RUN /usr/libexec/cont-setup
```
bparees: move the ADD above the first RUN and then merge this RUN into the existing RUN?  Saves a layer which isn't critical since we squash, but it makes the build faster too.

(You'll have to move the ENV up too since cont-setup depends on it, i don't think that has any negative ramifications though)

mfojtik: and please don't call it `cont-setup` but expand to `container-setup` :-) `cont` sounds very, very weird ;-)

hhorak:
test

bparees:
@hhorak this is a pretty big change that introduced a fair bit of complexity, I think it would help us if you outlined the use cases this solves that aren't addressed by the current implementation (or that are much harder to address w/ the current implementation).

I think we also need some docs to go with it that explain how and end user does extend/configure/etc this, since the stated intent is to make it more easily extensible.

thanks

@mfojtik fyi.

mfojtik:
+1 

In order to have this merged I would like to see documentation about how users can use this to extend the default behavior of the image (with few examples/use cases). You can add it to README.md or add `docs/` folder and just link to that. If we are going to adopt this, we will have to explain here https://docs.openshift.org/latest/using_images/db_images/mysql.html

hhorak:
I agree as well and will create something soon. Just to make clear, these changes doesn't invalidate any information from https://docs.openshift.org/latest/using_images/db_images/mysql.html, but I agree this text should be extended, together with README.md.

hhorak:
This is was just rebase to current master, readme is not yet extended..

hhorak:
Documentation is finally there and also few corresponding tests were added. Anybody able to take a look?

hhorak:
ping, any update?

bparees:
i'll let @mfojtik weigh in on that, he indicated you guys had discussed holding off on this.

mfojtik:
@hhorak this is in my post-PTO queue, sorry for the delay

README.md
```diff
@@ -122,6 +122,172 @@ your container. This will allow you to login to the root account remotely. Local
...
+* by volume mounting files during running the Docker image
+
+
+Creating a new layered Docker image on top of this Docker image
+---------------------------------------------------------------
+In order to use this Docker image as base in other layered Docker image, specify name of this one as base and then add files or run commands as you need, like this:
+```
+FROM <this-image>
+ADD mystuff /usr/bin/
+RUN /usr/bin/mystuff
```
mfojtik: i would mention that the entrypoint of the image will remain unchanged if you perform these changes

README.md
```diff
@@ -122,6 +122,172 @@ your container. This will allow you to login to the root account remotely. Local
...
+Example of the layered Dockerfile:
+
+```
+FROM <this-image>
+ADD myconfig.cnf /etc/my.cnf.d/myconfig.cnf
+```
+
+Notice: Read order of configuration files in `/etc/my.cnf.d/` directory is not specified.
+
+### Adjusting database inicialization
```
mfojtik: initialization

README.md
```diff
@@ -122,6 +122,172 @@ your container. This will allow you to login to the root account remotely. Local
...
+
+Initialization of mysql database during first start of this image is devided into several files in several directories. All directories include files that are read in alphabetical order, and thus file names are usually prefixed with a number.
```
mfojtik: divided

README.md
```diff
@@ -122,6 +122,172 @@ your container. This will allow you to login to the root account remotely. Local
...
+
+Concretly there are these directories within the Docker container:
+
+* `/usr/share/container-layer/mysql/pre-init/` -- this directory includes shell scripts with extention `.sh` that will be sourced and thus executed before every inicialization of container (no matter whether the data directory is empty or not). At the time these scripts are executed, the daemon is not yet running. Example of script in this directory is a script that checks arguments specified by user.
```
mfojtik: what does it mean '/usr/share/container-layer/**mysql**' ? Does it mean I can give names to layers?

mfojtik: This is already being handled in OpenShift using the lifecycle hooks and we don't want people to change the images to perform this initialization. You can execute a bash script(s) using the lifecycle hook directly in the container that runs the image. See: https://docs.openshift.org/latest/dev_guide/deployments.html#lifecycle-hooks

README.md
```diff
@@ -122,6 +122,172 @@ your container. This will allow you to login to the root account remotely. Local
...
+* `/usr/share/container-layer/mysql/post-init/` -- this directory includes shell scripts with extention `.sh` that will be sourced and thus executed before the first start of container (i.e. in case the container is started with empty data directory). At the time these scripts are executed, the daemon runs on localhost only and `mysql` utility should use arguments stored in `$mysql_flags` environment variable. This directory usually includes scripts for creating new databases, users or initialize the database according to your needs.
```
mfojtik: Same here, this is already being solved by `post` deployment lifecycle hook.

README.md
```diff
@@ -122,6 +122,172 @@ your container. This will allow you to login to the root account remotely. Local
...
+* `/usr/share/container-layer/mysql/pre-start/` -- this directory includes shell scripts with extention `.sh` that will be sourced and thus executed before every start of container (no matter whether the data directory is empty or not). At the time these scripts are executed, the daemon runs on localhost only and `mysql` utility should use arguments stored in `$mysql_flags` environment variable. Example of script in this directory is a script that allows to change password.
```
mfojtik: I see this a bit over-engineered. Can't this be done by `post-init` ? And leave it up to whoever is customizing the image if he cares about whether the data are initialized or not.

README.md
```diff
@@ -122,6 +122,172 @@ your container. This will allow you to login to the root account remotely. Local
...
+* `/usr/share/container-layer/mysql/include/` -- this directory includes shell scripts with extention `.sh` that will be sourced and thus executed in the beginning of the commands like run-mysqld. Files in this directory contain usually functions or variables definition that are supposed to be used in other scripts used in the image.
```
mfojtik: I would rename this to "common" but I'm not 100% on board with this idea ;-)

README.md
```diff
@@ -122,6 +122,172 @@ your container. This will allow you to login to the root account remotely. Local
...
+* `/usr/share/container-layer/mysql/usage/` -- this directory includes text files with extention `.txt` that will be printed as usage message for the container in case wrong arguments are given. Shell variables in these files will be expanded using `envsubst` utility.
```
mfojtik: I would avoid using file extensions as something that is affecting the functionality. Just say that this directory contains "text" files.

Also I don't think that this is needed at all. If I change usage of the image, then I most likely want to change the original usage file and not adding a new one to this directory and removing old one (which is essentially the same thing as just modifying the original usage ;-)

README.md
```diff
@@ -122,6 +122,172 @@ your container. This will allow you to login to the root account remotely. Local
...
+
+Generally, to perform some additional action according to your needs, put appropriate files into these directories in your Dockerfile. See other files in those directories for inspiration.
```
mfojtik: But you will have to rebuild the Docker image if you do, correct?

README.md
```diff
@@ -122,6 +122,172 @@ your container. This will allow you to login to the root account remotely. Local
...
+
+### Example of extending the Docker image by creating layered Dockerfile
+
+This example adds posibility to add a new user during database inicialization. The username and password is specified using `MYSQL_CONTENT_ADMIN_USER` and `MYSQL_CONTENT_ADMIN_PASSWORD` environment variables, provided when starting the container the first time.
+
+Since you might need to validate the input provided by user in environment variables, it is possible to do it by droping a shell script appropriate directory, that is sourced in the beginning of the daemon start.
```
mfojtik: I think that the validation should be done when you processing a template and providing the values for the environment variables. Personally, I think the Docker image should not perform any validations. If you craft the Docker image "mysql-1" to perform additional validations, but you also want to use the original image, you need to have 2 docker image in your storage, right?

And the fact that we're squashing the Docker images just makes this worst.

README.md
```diff
@@ -122,6 +122,172 @@ your container. This will allow you to login to the root account remotely. Local
...
+Finally, we want to extend the usage message that is printed in case wrong arguments are given by user. This is done by droping a text file into appropriate directory
+
+Content of the layered Dockerfile:
+```
+FROM <this-image>
+USER 0
+ADD 60_content_admin_create.sh /usr/share/container-layer/mysql/post-init/
+ADD 60_content_admin_password.sh /usr/share/container-layer/mysql/pre-start/
+ADD 60_verify_content_admin.sh /usr/share/container-layer/mysql/pre-init/
+ADD 60_usage_content_admin.txt /usr/share/container-layer/mysql/usage/
```
mfojtik: So to add a new script here and to make sure that I'm adding it into right 'position', I first have to `docker pull` the image I want to extended (given that I might be pulling already extended image that contains some customization scripts), then I have to docker run this image and manually go inspect the content of /usr/share/container-layer directories. Is that right?

README.md
```diff
@@ -122,6 +122,172 @@ your container. This will allow you to login to the root account remotely. Local
...
+USER 27
```
mfojtik: This instruction is invalid with default OpenShift SCC policy. We will run the image as random UID.

README.md
```diff
@@ -122,6 +122,172 @@ your container. This will allow you to login to the root account remotely. Local
...
+Content of the `60_content_admin_create.sh` file
+```
+mysql $mysql_flags <<EOSQL
+    CREATE USER '${MYSQL_CONTENT_ADMIN_USER}'@'%' IDENTIFIED BY '${MYSQL_CONTENT_ADMIN_PASSWORD}';
+    GRANT INSERT, SELECT, UPDATE, DELETE ON \`${MYSQL_DATABASE}\`.* TO '${MYSQL_USER}'@'%' ;
+    FLUSH PRIVILEGES ;
+EOSQL
+```
+
+Content of the `60_content_admin_password.sh` file
```
mfojtik: I like this way how you can structure the shell scripts. You can nicely plug in user creation, database creation and have them in separate files with documentation.

README.md
```diff
@@ -122,6 +122,172 @@ your container. This will allow you to login to the root account remotely. Local
...
+```
+Content admin user:
+  MYSQL_CONTENT_ADMIN_USER (max 16 characters)
+  MYSQL_CONTENT_ADMIN_PASSWORD
+```
+
+To build such a layered Dockerfile, just run `docker build .` in the directory, that includes all the files above. Built Docker image may be then run as:
+
+```
+docker run -ti --rm -e MYSQL_USER=user -e MYSQL_PASSWORD=pass1 -e MYSQL_DATABASE=db -e MYSQL_CONTENT_ADMIN_USER=contentuser -e MYSQL_CONTENT_ADMIN_PASSWORD=pass2 <image>
```
mfojtik: I would rather see OpenShift/Kubernetes template/pod definition here.

README.md
```diff
@@ -122,6 +122,172 @@ your container. This will allow you to login to the root account remotely. Local
...
+Example of such a feature is creation of one database with specified name and user that can work with that database. In order to remove this feature, the files need to be removed in the Dockerfile:
+
+```
+FROM <this-image>
+RUN rm -f /usr/share/container-layer/mysql/post-init/20-base-database.sh \
+/usr/share/container-layer/mysql/pre-start/10-passwords.sh \
+/usr/share/container-layer/mysql/pre-init/10-validate-base-variables.sh
+```
+
+Such an image won't require `MYSQL_USER`, `MYSQL_PASSWORD` and `MYSQL_DATABASE` environment variables specified.
```
mfojtik: And the only way how user will learn about it is to run the image and see the usage (means that you also have to provide one).

README.md
```diff
@@ -122,6 +122,172 @@ your container. This will allow you to login to the root account remotely. Local
...
+
+
+Extending the docker image by volume mounting files
+---------------------------------------------------
+
+This Docker image allows to adjust functionality even without creating the layered Dockerfile as described above. It is possible to use volume mounting using `-v` option (see `docker-run` man page for details).
+
+In a nutshell, all we need to do is adding the same files, as we added in examples above, just using volume mounting feature of the `docker utility.
```
mfojtik: you're missing `

README.md
```diff
@@ -122,6 +122,172 @@ your container. This will allow you to login to the root account remotely. Local
...
+
+### Example of adjusting the configuration of docker image by volume mounting files
+
+The simplest way of adding a new configuration file is adding it directly into `/etc/my.cnf.d` directory, like this.
+
+```
+docker run -d -v myconfig.cnf:/etc/my.cnf.d/myconfig.cnf <this_image>
```
mfojtik: You will have to know how is the image structured and where are the appropriate folders first to do this, right? I mean the config file you're mounting has to know about the layout of the filesystem of the given image.

Also this won't work on Kubernetes/OpenShift if you don't provide a volume mount that is already populated with the config files that another container produced or you created by hand and uploaded to the volume.

mfojtik:
@hhorak a summary of my comments (please don't take the comments wrong, just want to have all questions answered before we commit our self to support this mechanism).

* I do like the layout of /usr/share/container-layout. We can then nicely refactor our existing initialization mechanisms to be more user-friendly and readable.
* I don't like the complexity of the customization you're proposing. It means users have to learn how to "customize" Red Hat images which is not good. Users are already used to work with existing tools `docker build`.
* I do understand that all this is basically to support the `docker run` way to run things, which is not bad, but should not be our high priority. For example extending container by volume mounting configs is definitely not a way to go in Kubernetes/OpenShift (at least now). If you think, there is a high demand for this, we can add it to our roadmap, but it will have to go through "feature proposal" process.

I think we should start with removing functionality from this PR that is not absolutely necessary to make the initial refactoring of all images, namely:

* Use `/usr/share/container-scripts/[pre|post]-init` (no need to have extra folder for layers and will make the deploymentConfig lifecycle hooks more complex)
* Remove `pre-start` and `post-start` hooks (can be handled with above)

@bparees FYI

README.md
hhorak: `mysql` is used because we want some common scripts for more images be shared, so those will go into `/usr/share/container-layer/common/`. Then, those scripts might go into rpm once and then we can't let those scripts conflict between different packages.

hhorak: As for lifecycle hooks -- that's something we might use in OpenShift and k8s only, right? But we must have the image be usable even as alone, using `atomic run`, where we simply can't depend on k8s nor openshift features.

hhorak: Maybe I don't understand the lifecycle hooks correctly, so I may just stop by to consult this magic feature.

README.md
hhorak: `pre-init` is the place where you validate variables that are used during initialization and `post-init`. You can't just shuffle with those, because we don't want the initialization to even start in case validation fails.

README.md
hhorak: `common` works for me.. btw. I don't like the idea much either, but we want to share some functions and some code.. so we need some place where to place the shared code. Any better idea how to share it?

README.md
hhorak: Well, changing original file always implies that you must pay attention what you need to change and reflect the non-changed information in your copy. In case you change everything, you're way might be easier.. But I more counted with scenery, where you don't want to change everything, but just add something to existing feature set.

README.md
hhorak: Yes, will add that information there.

README.md
hhorak: Not sure I'm following, do you say users shouldn't adjust the image functionality by building additional layered image and should just use the existing one and provide the missing features when running the container?

README.md
hhorak: I don't understand why you'd need to run & inspect the image. You can definitely do it, but reading the documentation should also be enough -- you'll just follow the example and put your scripts into correct position in the Dockerfile.

README.md
hhorak: You can always run the image as random UID, but some UID is default (when not using `-u`). This is that default value. It is the same as what is used in original mysql Dockerfile.

README.md
hhorak: But we're supposed to deliver image usable even without those technologies. We're not blocking you from doing it with OpenShift/Kubernetes template/pod definition but we need something different.

README.md
hhorak: Right, we should also remove the part of the usage message that suggests `MYSQL_USER`, `MYSQL_PASSWORD` and `MYSQL_DATABASE` environment variables.

README.md
hhorak: thx, all the fixes for typos are already on the way.

README.md
hhorak: Yes, the path to config file is part of the image's API, same as path for the data directory is. In OpenShift you don't want to change configuration this way, do you? There is nice UI for that. Again, this necessary for use cases when somebody needs to run mysql database on Atomic host for example.

README.md
mfojtik: @hhorak I mean a case when you're going to build up on something that somebody published with his customization scripts.

hhorak: ah, so when creating a layered image on top of layered image on top of mysql image? well, I don't think this use case is something 80% of our users will do, but yes, in this case someone would have to inspect the image probably or the first customization would have to provide sane documentation about how to further extend the image.

hhorak:
I've reacted on most of the comments above, except idea of removing container-volume and just using one directory -- container-scripts.. I'm fine with that and will try to do the change hopefully tomorrow.. but I don't think we can get rid of pre-start and post-start hooks, they are just used in different conditions.. (more in comment above).

hhorak:
(the comment I refer is already hidden, so c&p here)
pre-init is the place where you validate variables that are used during initialization and post-init. You can't just shuffle with those, because we don't want the initialization to even start in case validation fails.

README.md
mfojtik: @hhorak that is what the layering grants you. I can imagine an use case where somebody builds an mysql image that will add some performance optimizations on top of the original mysql image (layer #1) and then publish it on Dockerhub. Other user find this image and add more configuration to it (like create different users/etc), which is layer #2. Now you have couple problems:

1) How the second user know what modifications were done in layer #1? He has to go pull the image and inspect the configuration files manually
2) What happens when the user supplying layer #1 change something that will break layer #2? I guess the user again has to go investigate
3) The second user will have to rebuilt the image everytime 1) layer #1 is updated, 2) base mysql image is updated. 

I'm just trying to highlight possible problems users might have by extending the image by adding more customization scripts into it.

README.md
mfojtik: @hhorak taking my original comment back. @bparees told me that we will keep this in image. I thought we are going to have something random here. nvmd.

README.md
mfojtik: @hhorak I know but ultimately this image should work in both environments. For that we should give people docs that covers both.

hhorak:
Closing in favor of https://github.com/openshift/mysql/pull/100 that should implement what we've agreed with here, in a simpler way.

PR #70, opened 2015-07-23
Lower the buffer pool size for InnoDB when running tests

mnagy:
This should prevent out-of-memory errors when running tests.

@bparees @jhadvig this should fix the push job.


bparees:
awesome, thanks @mnagy !

PR #71, opened 2015-07-28
Use rhscl namespace and naming conventions for mysql-5.6 Docker image

hhorak:
SSIA


bparees:
[test] (not that i'm worried about it, just checking something in jenkins)

PR #78, opened 2015-08-04
Remove shifts

mnagy:
The slave and master replicas do not expect commands. This caused replication containers to not run, as found by QE.

@mfojtik @mdshuai


jhadvig:
LGTM

mnagy:
@bparees fyi, I think this bug was responsible for our push job issues. When testing the replication, the innodb buffer flag was "shifted out" and not used in the end. The presence of that flag also hid the bug from the tests and that is why QE found it first, because they are testing the whole replica template example.

PR #79, opened 2015-08-04
WIP: Fix replication and add OpenShift tests

mnagy:
This script fixes the MySQL replication by creating the $MYSQL_DATABASE on the slave, so the bin log will be able to catch on. It also adds a test script that runs OpenShift inside of a Docker container and tests the replication scenario (our example template). It also makes use of a nfs-server image that I created in order to provide Persistent Volumes.

Please ignore `test/run` and `test/common.sh`, I'm going to refactor and change them later, right now I just hacked it quickly so I'll see how Jenkins will run the tests.

### DIFF
--- a/5.5/contrib/common.sh
+++ b/5.5/contrib/common.sh
@@ -86,10 +86,10 @@ function initialize_database() {
   mysql_install_db --datadir=$MYSQL_DATADIR
   start_local_mysql "$@"
 
-  [ -v MYSQL_DISABLE_CREATE_DB ] && return
-
   mysqladmin $admin_flags -f drop test
   mysqladmin $admin_flags create "${MYSQL_DATABASE}"
+
+  [ -v MYSQL_DISABLE_CREATE_DB ] && return
 
 mysql $mysql_flags <<EOSQL
     CREATE USER '${MYSQL_USER}'@'%' IDENTIFIED BY '${MYSQL_PASSWORD}';
--- a/5.5/test/test_replication.sh
+++ b/5.5/test/test_replication.sh
@@ -74,6 +74,10 @@ function start_openshift() {
       --loglevel=5
 }
 
+function dump_logs() {
+  docker logs openshift-origin
+}
+
 #
 # Helper functions to run commands inside of the OpenShift Docker container.
 #
@@ -112,12 +116,8 @@ cat examples/replica/mysql_replica.json | run_interactive "oc process -f - | oc 
 
 # Wait until master and slave are up.
 #set +x
-sleep 30
-dig "mysql-master.replication.svc.cluster.local"
-dig "@${HOST_DOCKER_IP}" "mysql-master.replication.svc.cluster.local"
-dig "mysql-master.replication.svc.cluster.local."
-dig "@${HOST_DOCKER_IP}" "mysql-master.replication.svc.cluster.local."
 echo "Waiting for MySQL Master and Replica to come online"
+trap dump_logs ERR
 wait_for_url_timed "mysql-master.replication.svc.cluster.local:3306" "" 5*TIME_MIN >/dev/null
 wait_for_url_timed "mysql-slave.replication.svc.cluster.local:3306" "" 1*TIME_MIN >/dev/null
 set -x
--- a/5.5/test/test_replication.sh
+++ b/5.5/test/test_replication.sh
@@ -1,0 +1,137 @@
+#!/bin/bash
+
+set -eux
+
+TIME_SEC=1000
+TIME_MIN=$((60 * $TIME_SEC))
+HOST_DOCKER_IP=$(ip -4 addr show docker0 | grep -Po 'inet \K[\d.]+')
+OPENSHIFT_CONFIG_DIR="/tmp/openshift-config"
+
+# time_now return the time since the epoch in millis
+function time_now()
+{
+  echo $(date +%s000)
+}
+
+# wait_for_url_timed attempts to access a url in order to
+# determine if it is available to service requests.
+#
+# $1 - The URL to check
+# $2 - Optional prefix to use when echoing a successful result
+# $3 - Optional maximum time to wait before giving up (Default: 10s)
+function wait_for_url_timed {
+  STARTTIME=$(date +%s)
+  url=$1
+  prefix=${2:-}
+  max_wait=${3:-10*TIME_SEC}
+  wait=0.2
+  expire=$(($(time_now) + $max_wait))
+  set +e
+  while [[ $(time_now) -lt $expire ]]; do
+    out=$(curl -k --max-time 2 -fs $url 2>/dev/null)
+    if [ $? -eq 0 ]; then
+      set -e
+      echo ${prefix}${out}
+      ENDTIME=$(date +%s)
+      echo "[INFO] Success accessing '$url' after $(($ENDTIME - $STARTTIME)) seconds"
+      return 0
+    fi
+    sleep $wait
+  done
+  echo "ERROR: gave up waiting for $url"
+  set -e
+  return 1
+}
+
+function setup_dns() {
+  # Make ourselves the default resolver. This is needes so FQDNs such as
+  # "pod-name.namespace.svc.cluster.local" can be resolved.
+  if ! grep -q $HOST_DOCKER_IP /etc/resolv.conf; then
+    sed -i "1inameserver $HOST_DOCKER_IP" /etc/resolv.conf
+  fi
+
+  sudo rm -rf $OPENSHIFT_CONFIG_DIR
+  # Generate openshift config file and edit the node config with DNS pointing to us.
+  docker run --rm -ti --privileged --net=host \
+    -v $OPENSHIFT_CONFIG_DIR:/config \
+    openshift/origin start --write-config=/config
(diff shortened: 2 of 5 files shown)


5.5/contrib/common.sh
```diff
@@ -86,11 +86,11 @@ function initialize_database() {
   mysql_install_db --datadir=$MYSQL_DATADIR
   start_local_mysql "$@"
 
-  [ -v MYSQL_DISABLE_CREATE_DB ] && return
-
   mysqladmin $admin_flags -f drop test
   mysqladmin $admin_flags create "${MYSQL_DATABASE}"
 
+  [ -v MYSQL_DISABLE_CREATE_DB ] && return
```
bparees: you're creating the DB before checking if create db is disabled?

bparees: you're creating the DB before checking if create db is disabled?

5.5/test/test_replication.sh
```diff
@@ -0,0 +1,145 @@
...
+EOF
+
+sleep 2
+
+# FIXME: Should log in as normal user, doesn't work right now.
+# Verify that values are replicated in slave.
+SLAVE_POD_NAME=$(run "oc get pods | grep mysql-slave | cut -f 1 -d ' '" | head -n 1)
+run "oc exec $SLAVE_POD_NAME -- /bin/bash -c 'mysql -uroot ${MYSQL_DATABASE} -e \"SELECT * FROM tbl;\"'" | grep -q foo1
+
+echo "All tests finished successfully"
```
bparees: cleanup/shutdown?

5.5/test/test_replication.sh
```diff
@@ -0,0 +1,145 @@
...
+  sudo rm -rf $OPENSHIFT_CONFIG_DIR
+  # Generate openshift config file and edit the node config with DNS pointing to us.
+  docker run --rm -i --privileged --net=host \
+    -v $OPENSHIFT_CONFIG_DIR:/config \
+    openshift/origin start --write-config=/config
+  sudo sed -i "s/dnsIP: .*/dnsIP: $HOST_DOCKER_IP/" $OPENSHIFT_CONFIG_DIR/$OPENSHIFT_NODE_CONFIG
+}
+
+# Start openshift. We're using local generated config file that has DNS updated.
+function start_openshift() {
```
bparees: ugh.  this is good work but also ugh (that we're adding all this logic to start/configure openshift in yet another location)

5.5/test/test_replication.sh
mnagy: Hmm, this is going to be a pickle. I usually cleaned up with `docker stop $(docker ps -q) ; docker rm -f $(docker ps -qa)`. It is difficult to shut down containers that OS launched, because we do not know their name and they keep running after stopping OS. Only thing I see in the openshift-origin container is the following line:
```
2015-08-04T08:03:56.902211104Z 2015/08/04 08:03:56 received terminated signal, shutting down...
```
Maybe this is a bug and OS should properly shutdown everything?

5.5/contrib/common.sh
mnagy: Yeah, I guess the variable's name no longer makes sense. Will rework this part as I start to figure out how to enable user replication, since the code is that code is there as well.

5.5/test/test_replication.sh
mnagy: Hmm, maybe going to try `oc delete all --all`.

mfojtik:
@bparees @mnagy I don't like where this is going. I mean it is great stuff and testing this image in OpenShift is a good thing, but I don't like we use bash scripts for this and I'm afraid that we end up re-creating test-cmd.sh here.

I would much rather see these tests as 'extended tests' in Origin and have them running here as part of our Jenkins job.

mfojtik:
@bparees @mnagy also if this will land, I would expect similar tests for mongodb and postgresql and then we came across the problem of duplicating the openshift server startup/etc... I would like to go one step back here and brainstorm about how we can make our testing for images better. I think @rhcarvalho  was also exploring this area.

mnagy:
@mfojtik Putting them all into one repository..

mfojtik:
@mnagy putting what into one repository? The bash library you wrote for bootstraping the server? I think we already have this here: https://github.com/mfojtik/origin/blob/extended-tests/hack/test-extended/default.sh

Is there a reason to duplicate this effort? Is there a reason to write all these tests in bash and not use the extended framework that was build to support these test-cases and share the common helpers?

mnagy:
@mfojtik I meant put all images into one repository.

The reason why the effort is duplicated is that we want to run the tests for PRs. We do not want to check out and build OpenShift every time we want to test an image, because it takes time and if OpenShift is broken, our PR tester will be as well. That is why I'm running OpenShift in Docker. Same goes for OpenShift - you test on a released MySQL image.

The "bash library" by itself is irrelevant to me, I don't really care about it being written in go. I just didn't know what else to use to quickly test what I need and bash was the easiest.

mfojtik:
@mnagy I thought we refused the idea of having all images in a single repository.

We don't need to build OpenShift (and you are not building it as well AFAIK). You just need to have OpenShift binaries available in order to start the extended test server (in whatever configuration).
There is no reason why you would want to run the OpenShift in Docker of you can just run in via './hack/test-extended/default.sh'.

@bparees I'm worried that the time spend here to write these tests is not worth if we are going to switch to use extended tests at some point (they will get merged today). I would rather want to see us focusing on moving to use them than writing 'quick && dirty' bash script we are going to get rid if in near future.

rhcarvalho:
@mfojtik it is very likely that the framework you've been working on is already light-years ahead of what I had. Also good that it is aligned with upstream. I believe 'your' framework will be able to run tests in parallel?

However, for testing images we might need something that can be decoupled from OS -- so that we can run tests locally during development and as part of PRs. @mfojtik that's where the tester image we discussed like a month ago would fit in.

Agree with @mfojtik and @bparees that replicating efforts on writing a test framework for this is not good. But I acknowledge the pain @mnagy is exposing here. As much as there is overlapping with other testing efforts, there's also some particular needs that I'm unsure the new extended tests in OS will be able to address.

Conclusion:

> I would like to go one step back here and brainstorm about how we can make our testing for images better.

+1

And, last but not least, I feel we're missing a clear picture of what "replication" in our DB images are meant for. Would like more thinking / clarification on that area too. The sort of apps I'd write and consider MongoDB or MySQL and their replication / sharding capabilities vary quite a bit, they're not drop-in replacements for each other, and OSv3 allow us to build even more sophisticated distributed systems where e.g. we use both MongoDB, MySQL and what not in an integrated fashion + replication and sharding on top of it. Think MongoDB for fresh live data + MySQL for historical data and a background job streaming data from one to the other.

mnagy:
@mfojtik I'm all for if we'll be able to test everything we need in a PR jenkins job, reasonably fast. I'd like to see an example implementation of a test for some of our DB image. .Any idea how long would it take to whip up one? We should discuss together with @bparees on how to do tests in image repositories.

mfojtik:
@mnagy https://github.com/mfojtik/origin/blob/extended-tests/test/extended/default_mysql_test.go this is an example of extended test.

@rhcarvalho yes, it will allow you to execute all tests in parallel (orchestrated by Ginko). In fact kubernetes e2e framework is already doing that for kubernetes. So in the end, it might look that using extended tests from OpenShift might slow down testing of PR, but in fact it might make the much faster.

5.5/test/test_replication.sh
```diff
@@ -0,0 +1,169 @@
...
+
+sleep 2
+
+# Verify that values are replicated in slave.
+SLAVE_POD_NAME=$(get_pod_name mysql-slave)
+# FIXME: Should log in as normal user, doesn't work right now.
+run "oc exec $SLAVE_POD_NAME -- /bin/bash -c 'mysql -uroot ${MYSQL_DATABASE} -e \"SELECT * FROM tbl;\"'" | grep -q foo1
+
+echo "All tests finished successfully, cleaning up"
+run "oc delete all --all"
```
bparees: `docker ps | awk 'index($NF,"k8s_")==1 { print $1 }' | xargs -l -r docker stop` should kill all containers launched by openshift.  they are all prefixed with k8s_ i believe.

bparees: (but if delete all works, that's certainly cleaner)

bparees: still need to stop openshift itself though, right?

rhcarvalho: `oc delete all --all` works, but operates on a single project/namespace, need to run it for every namespace (including 'default') -- that will stop containers, but not stop OS.

bparees:
so what i think we should be doing here is writing extended tests in origin that cover this scenario, and then invoking those specific extended tests from this repo's test job.

pretty sure that's what @mfojtik said and also what we discussed in scrum monday or tuesday.

that solves all the issues i see in that:
1) mysql PRs don't get merged w/o passing replication tests
2) replication tests can reuse common function in the extended test framework

am i missing something?

mnagy:
@bparees We agreed with @mfojtik yesterday that we would download released binaries of OpenShift so we do not need to compile them ourselves and to make sure that commits to OS master won't break image PRs.

bparees:
@mnagy sounds good.

bparees:
@mnagy is this still actively a WIP?

mnagy:
@bparees not really, since I'm reworking the tests for origin extended tests. Closing.

PR #80, opened 2015-08-07
Change remove_tmp_dockerfile

stevekuznetsov:
The `-z` flag returns `true` if a string length is 0. `[ ! -z "$var" ]` is equivalent to `[ -n "$var" ]` or `[ "$string" ]` and in either case is not testing what is necessary here - because the string in question, `"${DOCKERFILE_PATH}.version"`, will *never* be of length 0, this `if` statement is not doing much.

Furthermore, this trap only removes the last `Dockerfile` used for `docker build`, and when running a  naked `make` command, multiple `docker build`s are run.


mfojtik:
:+1: good catch! LGTM

stevekuznetsov:
@mfojtik Let me know if you think I missed any!

PR #83, opened 2015-08-25
Add support for atomic command

hhorak:
Atomic command is standard way to run containers on Atomic host. This adds support for running this image on such a system using `atomic install/uninstall/run`, as described at https://github.com/projectatomic/atomic/.

@omron93 @hanzz have you got to something similar to that?


5.5/Dockerfile
```diff
@@ -20,10 +20,23 @@ LABEL io.k8s.description="MySQL is a multi-user, multi-threaded SQL database ser
...
+      --ipc=host --pid=host -e HOST=/host -e LOGDIR=\${LOGDIR} -e CONFDIR=\${CONFDIR} \
+      -e DATADIR=\${DATADIR} -e IMAGE=\${IMAGE} -e NAME=\${NAME} -e OPT1 -e OPT2 -e OPT3 \
+      \${OPT1} \${IMAGE} /usr/share/container-layer/mysql/atomic/install.sh
+LABEL UNINSTALL docker run -t -i --rm --privileged -u 0:0 -v /:/host --net=host \
+      --ipc=host --pid=host -e HOST=/host -e LOGDIR=\${LOGDIR} -e CONFDIR=\${CONFDIR} \
+      -e DATADIR=\${DATADIR} -e IMAGE=\${IMAGE} -e NAME=\${NAME} -e OPT1 -e OPT2 -e OPT3 \
+      \${OPT1} \${IMAGE} /usr/share/container-layer/mysql/atomic/uninstall.sh
+LABEL RUN docker run -t -i --rm --name=\${NAME} -e OPT1 -e OPT2 -e OPT3 \${OPT2} \
+      \${IMAGE}
+
```
bparees: you can do these all as a single layer to save build time.

bparees: in fact combine them into the label statement above.

5.5/Dockerfile.rhel7
```diff
@@ -26,10 +26,23 @@ LABEL BZComponent="mysql55-docker" \
...
+      --ipc=host --pid=host -e HOST=/host -e LOGDIR=\${LOGDIR} -e CONFDIR=\${CONFDIR} \
+      -e DATADIR=\${DATADIR} -e IMAGE=\${IMAGE} -e NAME=\${NAME} -e OPT1 -e OPT2 -e OPT3 \
+      \${OPT1} \${IMAGE} /usr/share/container-layer/mysql/atomic/install.sh
+LABEL UNINSTALL docker run -t -i --rm --privileged -u 0:0 -v /:/host --net=host \
+      --ipc=host --pid=host -e HOST=/host -e LOGDIR=\${LOGDIR} -e CONFDIR=\${CONFDIR} \
+      -e DATADIR=\${DATADIR} -e IMAGE=\${IMAGE} -e NAME=\${NAME} -e OPT1 -e OPT2 -e OPT3 \
+      \${OPT1} \${IMAGE} /usr/share/container-layer/mysql/atomic/uninstall.sh
+LABEL RUN docker run -t -i --rm --name=\${NAME} -e OPT1 -e OPT2 -e OPT3 \${OPT2} \
+      \${IMAGE}
+
```
bparees: again merge into the LABEL command above.

5.6/Dockerfile
```diff
@@ -20,10 +20,23 @@ LABEL io.k8s.description="MySQL is a multi-user, multi-threaded SQL database ser
...
+      --ipc=host --pid=host -e HOST=/host -e LOGDIR=\${LOGDIR} -e CONFDIR=\${CONFDIR} \
+      -e DATADIR=\${DATADIR} -e IMAGE=\${IMAGE} -e NAME=\${NAME} -e OPT1 -e OPT2 -e OPT3 \
+      \${OPT1} \${IMAGE} /usr/share/container-layer/mysql/atomic/install.sh
+LABEL UNINSTALL docker run -t -i --rm --privileged -u 0:0 -v /:/host --net=host \
+      --ipc=host --pid=host -e HOST=/host -e LOGDIR=\${LOGDIR} -e CONFDIR=\${CONFDIR} \
+      -e DATADIR=\${DATADIR} -e IMAGE=\${IMAGE} -e NAME=\${NAME} -e OPT1 -e OPT2 -e OPT3 \
+      \${OPT1} \${IMAGE} /usr/share/container-layer/mysql/atomic/uninstall.sh
+LABEL RUN docker run -t -i --rm --name=\${NAME} -e OPT1 -e OPT2 -e OPT3 \${OPT2} \
+      \${IMAGE}
+
```
bparees: merge

5.6/Dockerfile.rhel7
```diff
@@ -27,10 +27,23 @@ LABEL BZComponent="rh-mysql56-docker" \
...
+      --ipc=host --pid=host -e HOST=/host -e LOGDIR=\${LOGDIR} -e CONFDIR=\${CONFDIR} \
+      -e DATADIR=\${DATADIR} -e IMAGE=\${IMAGE} -e NAME=\${NAME} -e OPT1 -e OPT2 -e OPT3 \
+      \${OPT1} \${IMAGE} /usr/share/container-layer/mysql/atomic/install.sh
+LABEL UNINSTALL docker run -t -i --rm --privileged -u 0:0 -v /:/host --net=host \
+      --ipc=host --pid=host -e HOST=/host -e LOGDIR=\${LOGDIR} -e CONFDIR=\${CONFDIR} \
+      -e DATADIR=\${DATADIR} -e IMAGE=\${IMAGE} -e NAME=\${NAME} -e OPT1 -e OPT2 -e OPT3 \
+      \${OPT1} \${IMAGE} /usr/share/container-layer/mysql/atomic/uninstall.sh
+LABEL RUN docker run -t -i --rm --name=\${NAME} -e OPT1 -e OPT2 -e OPT3 \${OPT2} \
+      \${IMAGE}
+
```
bparees: merge

README.md
```diff
@@ -122,6 +122,36 @@ your container. This will allow you to login to the root account remotely. Local
 connections will still not require a password.
 
 
+Usage on Atomic host
+---------------------------------
+Systems derived from projectatomic.io usually include `atomic` command that is
```
bparees: s/include/include the/

README.md
```diff
@@ -122,6 +122,36 @@ your container. This will allow you to login to the root account remotely. Local
...
+used to run containers except other tasks.
```
bparees: except?

README.md
```diff
@@ -122,6 +122,36 @@ your container. This will allow you to login to the root account remotely. Local
...
+To install a new service `mysql1` based on this image on such a system, run:
+
+```
+$ atomic install -n mysql1 --opt2='-e MYSQL_USER=user` -e MYSQL_PASSWORD=secretpass -e MYSQL_DATABASE=db1 -p 3306:3306' THIS_IMAGE
+```
+
+Then to run the service, use the standard `systemctl` call:
+
+```
+$ systemctl start mysqld1.service
```
bparees: mysql1.service?  or mysqld1.service?

README.md
```diff
@@ -122,6 +122,36 @@ your container. This will allow you to login to the root account remotely. Local
...
 
+Usage on Atomic host
+---------------------------------
+Systems derived from projectatomic.io usually include `atomic` command that is
+used to run containers except other tasks.
+
+To install a new service `mysql1` based on this image on such a system, run:
+
+```
+$ atomic install -n mysql1 --opt2='-e MYSQL_USER=user` -e MYSQL_PASSWORD=secretpass -e MYSQL_DATABASE=db1 -p 3306:3306' THIS_IMAGE
```
bparees: "THIS_IMAGE" would be rhscl/mysql-56-centos7 for example?  might be better to provide a concrete example or something that at least looks more like a docker pull spec.

README.md
```diff
@@ -122,6 +122,36 @@ your container. This will allow you to login to the root account remotely. Local
...
+
+Then to run the service, use the standard `systemctl` call:
+
+```
+$ systemctl start mysqld1.service
+```
+
+In order to work with that service, you may either connect to exposed port 3306 or run this command to connect locally:
+```
+$ atomic run -n mysql1 THIS_IMAGE bash -c 'mysql'
```
bparees: same comment about THIS_IMAGE as a value.

README.md
```diff
@@ -122,6 +122,36 @@ your container. This will allow you to login to the root account remotely. Local
...
+```
+
+To stop and uninstall the mysql1 service, run:
+
+```
+$ systemctl stop mysql1.service
+$ atomic uninstall -n mysql1 THIS_IMAGE
```
bparees: same comment about THIS_IMAGE

bparees:
couple nits, mostly looks good to me.
[test]

README.md
hhorak: mysqld1 seems better, it is similar to main daemon name from mysql. So I'll fix the other occurrences.

README.md
hhorak: like "except other things" which can be used instead.

5.5/Dockerfile
hhorak: Ok, will do.

README.md
hhorak: All right, will fix..

5.5/Dockerfile
hhorak: Actually I'd rather just merge them together, but let all the Atomic labels as one LABEL instruction. The time saved might be 1s but the Dockerfile will be better readable.

hanzz:
@hhorak 

Petr Pisar had some opinion on my atomic install.sh script, which could be valid also for your install.sh, so I'm forwarding it here:

> Just from security point of view, the script will have to be much uglier. All
the variables should be quoted and many of them sanitized against leading
hyphens and docker or systemd special keywords.

hhorak:
@hanzz Thanks for pointing to that, will use quotes more :)

5.5/Dockerfile
```diff
@@ -20,10 +20,23 @@ LABEL io.k8s.description="MySQL is a multi-user, multi-threaded SQL database ser
       io.openshift.expose-services="3306:mysql" \
       io.openshift.tags="database,mysql,mysql55"
 
+# Support for atomic run/install/uninstall
+LABEL INSTALL="docker run -t -i --rm --privileged -u 0:0 -v /:/host --net=host \
```
mfojtik: this is incredibly ugly :-) @vpavlin are there any plans to make this crypto commands more human-friendly?

hhorak: Why ugly? This is transparent for the user, so why does it matter? Anyway, this may be raised in the atomic itself (https://github.com/projectatomic/atomic/), I don't think we can do anything differently in this image.

mfojtik: @hhorak just from looking at this, does it seem readable to you? It matters because we want users to build the images that will run on Atomic. Users will eventually copy the patterns we have here. My concern which this pattern is that many users will not really understand what this magic options are for and will end up copying this blindly.

I don't think we can solve this here and now, that is what I tagged @vpavlin to provide his thoughts.

mfojtik: cc @bparees

mfojtik: @hhorak also, I don't want to block this PR from being merged, just want to have discussion about these labels.

hhorak: I agree creating image for atomic should be easier and also should look easier. It is not that complicated in the end, but it looks complicated, which is not good.

mfojtik: @hhorak my concerns with this is:

1) this is not backward compatible and not namespaced, so everytime something changes in the Docker CLI, this will break terribly
2) A lot of repetition (why I need to write `docker run` three times if it is obvious this will run with docker)
3) What will happen if we start supporting rocket or another container provider? Will the INSTALL become `INSTALL.rocket=` ? 
4) Having this labels in one liner saves some time during a build, but parsing this just with eyes is very hard. I mean I completely missed RUN ;-)
5) The OPT1 OPT2... are totally cryptic and I have no idea what they are doing (need to go check atomic docs)
6) These labels will be copied to any other image that will layer on top of this image. If you change this label in the future, the 'upper' images will have to rebuild. There is no way to force rebuild of all images users will publish on Docker Hub, so this might be a huge problem in the future.
7. Why this has to be privileged? We spend a lot of effort to make these images run under arbitrary UID, now we are saying that they should run as root?

mfojtik: @rhatdan FYI

hanzz: @mfojtik , I think most of your questions are really questions for atomic developers. We are just trying to use what they provide. You are right that we should try to discuss our needs with them a bit and push them into "right" direction. @hhorak, maybe we should try more to discuss all of that with Atomic team and ask them for some solution for the problems @mfojtik pointed out?

> 7. Why this has to be privileged? We spend a lot of effort to make these images run under arbitrary UID, now we are saying that they should run as root?

As I understand Project Atomic, "atomic install" is designed to be run as root and it should execute single command which installs the docker image on the host system. Since single command is not enough for the creation of systemd service for the installed app or the volumes for the application, therefore the script is used.

That script is stored in the container itself and because the script must do all of that, it requires access to directories owned by root. Therefore it uses --privileged. I admit we could drop the rights in the script when calling assemble (or s2i in the future) and use the root only for things which strictly needs it.

Note that this is not something we have invented in RHSCL, it's documented in the Project Atomic [documentation](https://github.com/projectatomic/atomic/). Again, we could discuss with Project Atomic people if that's right or not, but for now we have decided to respect what Project Atomic offers and just use it...

mfojtik: @hanzz the OpenShift SCC defaults to run container as not root for very good reasons. I think the same reasons applies to the Atomic, @rhatdan ?

hanzz: > @hanzz the OpenShift SCC defaults to run container as not root for very good reasons. I think the same reasons applies to the Atomic, @rhatdan ?

And I agree with you here, the application in the container should run as non-root. The container installed by the "atomic install" command still defaults  to non-root user. It is only the installation process running as root for the reasons I've mentioned earlier. It is question for the Project Atomic developers if they see a way how to do that as non-root or if it's useful to do it that way.

mfojtik: @hanzz even the install is important as somebody can write malicious code to break into the host machine and run the commands as root there. I think we recently switched the OpenShift builds to also run as non-root @csrwng was working on that.

hanzz: Again, I don't disagree, but you can say that about RPM too and we are all using it to install SW as root. Anyway, let's stop here, since it's starting to be unconstructive now I think. Neither me or you can change the way how Atomic is working *right* now. Let's wait for someone from Atomic instead :).

vpavlin: @mfojtik I agree it's hard to figure out whole story behind the atomic labels, OTOH I don't agree it's that cryptic. Users will copy anything if they a) don't have reason to understand what they are doing b) can easily find good documentation.

There were ideas about making labels more declarative, but that would be again very problematic regarding backwards compatibility and ease of use. Single docker run command for each step makes sense to me (at least from what I can imagine as alternatives). You can easily inspect image and review the command, you can override it if needed...

Install (in most cases) needs to write to host, thus needs to run as privileged container. No, you don't need to always get rid of all namespaces, but sometimes you do. The analogy for this is `yum install` where, if you package some malicious code f.e. in `%post` script, you are in the same situation.

`OPTx` arguments are mess, I agree, though I have no idea how to fix them,,

mfojtik: @vpavlin so to summary:

* We don't know what is the story around Atomic labels (?) and it is fine for users to copy something they don't understand
* The current labels schema has some backward compatibility problems I highlighted here https://github.com/openshift/mysql/pull/83/files#r38512465 The problem with the 'docker command' is that is very hard to understand (at least for me, and I don't count as a regular user...)
* What you need to write to host here? You bind mount a directory where you write the database. This directory can be 0777 or in the worst case owned by arbitrary user (similar to OpenShift persistent volumes).
* If `OPTx` are mess and we all agree on that, then we should find a way to fix it.

hanzz: > OPTx arguments are mess, I agree, though I have no idea how to fix them,,

Just a stupid idea I've been thinking about for only few minutes, so it could be, well, stupid. Can't you just parse the label for the occurrence of $UNKNOWN variable, let the user use "atomic install --unknown=something" or "atomic install --set unknown=something" or something similar to "-e UNKNOWN=something" and replace $UNKNOWN with "something" the same way as you do for OPTx?

That way, the creator of docker image could at least decide the names of the variables.

mfojtik: @hanzz this sounds similar to openshift parameters, but I like the idea of substitution more than OPTx. At least you can more clearly describe the intent of that environment variable and give users clue about what is this env var going to do.

vpavlin: @mfojtik No, I didn't say we don't know the story. Just that it might not be obvious for the first glance. The story is to let user run the image properly without being forced to search for READMEs and How-tos. He just needs to know that there is an atomic command which is able to run images that support it.

Also, I didn't say it's ok that users copy random stuff and run it (Hey, have you tried to use Kubernetes and run examples without knowing what that lines mean?:) ). What I am saying is that users just do it that way and the only thing we can do is to provide them with accessible and straightforward documentation. 

Your questions about backward compatibilty are very good, but you are asking wrong people at wrong place. @rhatdan is trying to get attention to this  for some time already at https://github.com/projectatomic/ContainerApplicationGenericLabels

You need to write out the service file here. Or some configuration (f.e. in case of rsyslog image).

I like @hanzz's proposal/idea but again - this is a wrong place to solve it because not all interested people are following this particular pull request:-)

vpavlin: `$OPTx` is not only for environment variables, but for modifying the command overall. Let's say you have the `--verbose` switch which is off by default, but you can turn it on if there is `$OPTx` on the right place.

Images should work without these options being used, but they should let you tweak the behaviour of image in special cases (like debuging as I described above)

hhorak: The `$OPTx` in this image are used to get user/pass/dbname from user, which are required to create a container. I don't have any other way to get those data except `$OPTx`, do I?
@hanzz not sure I got your crazy idea correctly, but when I tried something like `atomic install --unknown ...`, then I got `/bin/atomic: unrecognized arguments: --unknown`...

soltysh: I totally agree with @mfojtik reading this is hurting my eyes. We've (OpenShift team) put a lot of effort to simplify how to create builder images, with this addition I'm more than 100% sure people will refuse to build those by themselves. One big questions that comes into my head is:
if these are directives for atomic command, can't these just point the necessary information (dirs, ports, etc.) you'll pass down to container engine (be that docker or rocket in the future) and atomic will know how to call docker/rkt run etc. Similarly to how we do with s2i build, we don't put entire logic into the builder image, we just put the necessary bits there to run the command, can't atomic do the same?

hanzz: > @hanzz not sure I got your crazy idea correctly, but when I tried something like atomic install --unknown ..., then I got /bin/atomic: unrecognized arguments: --unknown...

I was proposing the way how atomic could work in the future. It's not implemented like that right now :).

rhatdan:
Perhaps it would be better to create a BlueJeans meeting to discuss these issues.

I am open to ideas on how to fix these issues.  I think Vaclav has done a good idea describing what the goal was.  

BTW Current atomic will allow you to append content on to the end of the INSTALL/RUN/UNINSTALL lines

atomic install mysql --user=dwalsh --passwd=foobar

Should endup taking the INSTALL line and appending --user=dwalsh --passwd=foobar to the end.

mfojtik:
@rhatdan +1 (I will schedule something for next week, what day suits you best?)

mfojtik:
@hhorak can you rebase this?

hhorak:
@mfojtik as for one of you comments above.. "4) Having this labels in one liner saves some time during a build, but parsing this just with eyes is very hard. I mean I completely missed RUN ;-)" -- that was how it looked first time and i made one command from it after feedback here. If you prefer one command per LABEL, so as I do, shouldn't we return to the previous variant?

mfojtik:
No, I think it is fine (it makes the build faster)... We should work on
better solution than using Docker labels. Lets discuss that next week with
Dan.

On Wed, Sep 2, 2015 at 4:30 PM, Honza Horak <notifications@github.com>
wrote:

> @mfojtik <https://github.com/mfojtik> as for one of you comments above..
> "4) Having this labels in one liner saves some time during a build, but
> parsing this just with eyes is very hard. I mean I completely missed RUN
> ;-)" -- that was how it looked first time and i made one command from it
> after feedback here. If you prefer one command per LABEL, so as I do,
> shouldn't we return to the previous variant?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openshift/mysql/pull/83#issuecomment-137103095>.
>



-- 
-----------------------------------------------
Michal Fojtik <mi@mifo.sk <mfojtik@mifo.sk>>
Red Hat OpenShift, Engineering

rhatdan:
Monday is Holiday in US.  Tuesday After 10-15 EDT is open.  Also have scattered time Wednesday-Friday

5.6/atomic-contrib/include.sh
```diff
@@ -0,0 +1,12 @@
+# Check whether all required variables are set
```
mfojtik: @hhorak cab you rename this file to "variables.sh" or "vars.sh" ? As this file really just define variables used by other scripts.

mfojtik:
one renaming suggestion, otherwise LGTM, we will discuss the labels/format on Tuesday.

mnagy:
How would you run this image with atomic and specify a location for the data volume (/var/lib/mysql/data)?

mnagy:
OK, I see that you can use `OPT2`, I guess.

However, that's just weird and kind of defeats the purpose of making this generic for atomic. I can understand if the labels are meant to be used from shell, for example. But if you are running atomic {run,install,uninstall}, you should be able to use some "normal" metadata for guessing how to run it, no? For example, the metadata might say you need to mount the host rootfs, run privileged and execute a given script (or have a standard location for an atomic install script, so this would be unnecessary). You would get the VOLUME metadata for free.

hhorak:
@mnagy `atomic` should already contain some feature that allows to use the rest (or unknown) arguments in the docker command. I haven't investigated that yet, but that's what @rhatdan mentioned.

rhatdan:
OPT1, OPT2 and OPT3 should not be used unless the user/tester wants to override the atomic command.

We want 
```
atomic install mariadb
```

To just work.

If the user wanted to do something like

atomic install --opt1="--debug" mariadb

Then it could replace $OPT1 in the LABEL with "--debug" when it executes the command and potentially run the command in debug mode.  But OPT1,OPT2, OPT3 should not be part of the standard command that users execute.

> How would you run this image with atomic and specify a location for the data volume (/var/lib/mysql/data)?

The developer of mariadb would just specify the RUN LABEL to be 

```
LABEL RUN "docker run -v /var/lib/mysql:/var/lib/mysql ..."
```

You might have LABEL INSTALL command do something like running myslqd init to create content in the hosts /var/lib/mysql directory.

If you wanted to support multiple versions of mariadb you could do it based on $NAME

```
LABEL RUN "docker run -v /var/lib/\${NAME}/mysql:/var/lib/mysql ..."
```

Which would allow you to run multiple containers with different names.

```
atomic install --name=mariadb1 mariadb
atomic install --name=mariadb2 mariadb
atomic run --name=mariadb1 mariadb
atomic run --name=mariadb2 mariadb
```

mnagy:
@rhatdan would it be a normal use-case for atomic to have a DB on host? Would a user running `atomic run mariadb` expect that behaviour? If so, then the RUN label really should be mounting the volume.

The database is initialized automatically, when none is detected, so there should be no need for the install label to handle this.

rhatdan:
Well you would not have a /var/lib/mysql on the host unless you do an install.

hhorak:
@rhatdan Thanks for the feedback, I've tried to take all that into account when creating reworked version at https://github.com/openshift/mysql/pull/92, but there is one difference from your description (except the fact that some environment variables are compulsory in mysql image now, so running just `atomic install --name=mariadb2 mariadb` won't work and user is required to set MYSQL_USER and other stuff).

The difference is caused by https://github.com/projectatomic/atomic/issues/151 and especially by not using systemd unit for watching the container. How @vpavlin explained me it should work (correct me if I'm wrong) is that `atomic install` should create the container (`docker create`) and `atomic run` should just start that container:
```
atomic install -n mysql1 mysql-55-centos7
atomic run mysql1
atomic stop mysql1
atomic uninstall
```
(this workflow defines `RUN label as docker start ...`, so it makes it impossible to use `atomic run` without prior `atomic install`)

However, from what I see in atomic docs, it looks like the `atomic run` was supposed to be run also standalone (without `atomic install` first). Then we'd either need something like `atomic start` which would do what basically `docker start` does or we can point user not to use atomic to start already installed container and just advise this workflow:
```
atomic install -n mysql1 mysql-55-centos7
docker start mysql1
docker stop mysql1
atomic uninstall
```
Any advise which way is better here?

rhatdan:
atomic run will do a start if the container exists, if it does not exists it will execute the LABEL RUN

rhatdan:
I think these would all work.
```
atomic install -n mysql1 mysql-55-centos7
atomic run -n mysql1 mysql-55-centos7
atomic stop mysql1
atomic uninstall -n mysql1 mysql-55-centos7
```

hhorak:
I still struggle with some elegant implementation for `atomic run -n mysql1 mysql-55-centos7` being run without running `atomic install` before (which I understood is valid use case as well). The thing is the mysql image won't work unless we specify several `-e` options. Those options must go before image name in the docker call, otherwise those are not taken as docker options, but rather as CMD + args.

So, for example in case we call
```
atomic run -n mysql1 mysql-55-centos7 -e MYSQL_USER=myuser1
```
then what we get is
```
docker run  -n mysql1 mysql-55-centos7 -e MYSQL_USER=myuser1
```
while what we need is
```
docker run  -n mysql1 -e MYSQL_USER=myuser1 mysql-55-centos7
```
The easiest idea to fix this was to use script similar to install.sh, something like this:
```
$ cat run.sh
#!/bin/bash
docker run --name "${NAME}" "$@" "${IMAGE}"
```
and having RUN label like this:
```
LABEL RUN="docker run --rm \${IMAGE} cat /usr/share/container-layer/mysql/atomic/run.sh | bash -s --"
```
but I got the following error:
```
$ atomic run  -n myc4 mysql-55-rhel7 -p 3306:3306 -e MYSQL_USER=user -e MYSQL_PASSWORD=secretpass -e MYSQL_DATABASE=db1
Container 'myc4' must be running before executing a command into it.
Execute the following to create the container:
/usr/bin/atomic run --name myc4 mysql-55-rhel7
```
Any idea?

rhatdan:
This should probably be in the label.
-p 3306:3306

Why not have the User name and password passed after the command on the run, they will get appended as options on the command.

$ atomic run  -n myc4 mysql-55-rhel7 --user=user --password=secretpass --db=db1

rhatdan:
Then the setup script could use these parameters or prompt for them if they are not given.  

BTW this is the goal of nulecule/atomicapp to handle problems like this.

@vpavlin @markllama

hhorak:
> Why not have the User name and password passed after the command on the run, they will get appended as options on the command.

Now I see why it didn't work for me, I simply had too old version of `atomic`..

hhorak:
Closing this one in favor of https://github.com/openshift/mysql/pull/92, the same feature is gonna be done by that new PR.

PR #85, opened 2015-08-25
Replicate mysql database, fix replication

mnagy:
This should fix replication. Will add extended tests to origin.


5.5/contrib/common.sh
```diff
@@ -134,23 +146,19 @@ function wait_for_mysql_master() {
...
 # endpoints. Each endpoint is a MySQL container with the 'master' MySQL running.
 function mysql_master_addr() {
-  local service_name=${MYSQL_MASTER_SERVICE_NAME:-mysql-master}
-  local endpoints=$(dig ${service_name} A +search +short 2>/dev/null)
-  # FIXME: This is for debugging (docker run)
-  if [ -v MYSQL_MASTER_IP ]; then
-    endpoints=${MYSQL_MASTER_IP-}
-  fi
-  echo -n "$(echo $endpoints | cut -d ' ' -f 1)"
+  echo -n "${MYSQL_MASTER_SERVICE_NAME:-mysql-master}"
```
bparees: if that's all this is doing why do we need a function anymore?

5.5/contrib/my-master.cnf.template
```diff
@@ -2,6 +2,7 @@
 
 server-id     = ${MYSQL_SERVER_ID}
 log_bin       = ${MYSQL_DATADIR}/mysql-bin.log
+binlog_do_db  = mysql
 binlog_do_db  = ${MYSQL_DATABASE}
```
bparees: ?  aren't these overwriting each other?

5.5/contrib/my-slave.cnf.template
```diff
@@ -3,6 +3,7 @@
 server-id     = ${MYSQL_SERVER_ID}
 log_bin       = ${MYSQL_DATADIR}/mysql-bin.log
 relay-log     = ${MYSQL_DATADIR}/mysql-relay-bin.log
+binlog_do_db  = mysql
 binlog_do_db  = ${MYSQL_DATABASE}
```
bparees: same...  how are you setting the same key twice?

mnagy: No overwriting, this is how it's supposed to be. A database name can apparently contain a comma, that's the reason you need two options. See the red warning here: https://dev.mysql.com/doc/refman/5.5/en/replication-options-binary-log.html#option_mysqld_binlog-do-db

5.5/contrib/common.sh
mnagy: I guess it was partly laziness and partly to keep it in common.sh. But I now I think I'll get rid of the function. Will rework.

5.5/contrib/my-slave.cnf.template
bparees: got it, thanks.

bparees:
[test]

bparees:
ugh, sorry, mixed up my PRs and merged this one on you.  it did pass tests, so it should be ok but you'll have to do that one rework as a new PR.
:(

PR #86, opened 2015-08-26
Don't look up master address in DNS, use hostname directly

mnagy:
@bparees

### DIFF
--- a/5.5/run-mysqld-slave.sh
+++ b/5.5/run-mysqld-slave.sh
@@ -26,7 +26,7 @@ initialize_database "$@"
 wait_for_mysql_master
 
 # Get binlog file and position from master
-STATUS_INFO=$(mysql --host "$(mysql_master_addr)" "-u${MYSQL_MASTER_USER}" "-p${MYSQL_MASTER_PASSWORD}" replication -e 'SELECT File, Position from replication\G')
+STATUS_INFO=$(mysql --host "$MYSQL_MASTER_SERVICE_NAME" "-u${MYSQL_MASTER_USER}" "-p${MYSQL_MASTER_PASSWORD}" replication -e 'SELECT File, Position from replication\G')
 BINLOG_POSITION=$(echo "$STATUS_INFO" | grep 'Position:' | head -n 1 | sed -e 's/^\s*Position: //')
 BINLOG_FILE=$(echo "$STATUS_INFO" | grep 'File:' | head -n 1 | sed -e 's/^\s*File: //')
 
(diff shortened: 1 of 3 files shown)


mfojtik:
LGTM

5.5/run-mysqld-slave.sh
```diff
@@ -6,6 +6,7 @@ source ${HOME}/common.sh
 set -eu
 
 export MYSQL_RUNNING_AS_SLAVE=1
+export MYSQL_MASTER_ADDR=${MYSQL_MASTER_ADDR:-mysql-master}
```
bparees: what is setting this when running in openshift?  did you mean $MYSQL_SERVICE_NAME?  but even that isn't a resolveable ip or DNS hostname....

mfojtik: this is a head less service, SkyDNS provides this hostname

bparees: what provides the MYSQL_MASTER_ADDR env variable?

and are you saying "mysql-master" resolves to something via DNS?  what domain is it searching?

mfojtik: @bparees I think it is searching the current namespace for the "mysql-master" service. The variable is not provided, it is there so you can override it when you do bash testing with docker (where you don't have the headless service available)

bparees: since it's just a dns lookup for "mysql-master" i don't see how it would know to specifically search the current namespace unless that's part of the /etc/resolv.conf that's mounted into the pod?

the fact that it's not provided means this image is dependent on the service existing with exactly that name..that seems bad.  it used to be a variable based on the service name......

mfojtik: @bparees +1 we should use the service name instead of hard coding 'mysql-master' here.

mfojtik: @bparees resolving the DNS in namespace is magic to me... I remember Clayton was explain this to us once but I can't remember the technical details. From what I remember is that it is safe to assume to DNS lookup for `mysql-master` will give you the IP of the service running in the current namespace.

If you want to be more precise, you can also include the namespace name in the DNS query (mysql-master.NAMESPACE) but I don't think we have NAMESPACE variable available in the container.

mfojtik: OK, you can use https://github.com/kubernetes/kubernetes/blob/871ca7a05f3aeb0f2324b1d424ae67ada8425486/docs/user-guide/downward-api.md to get the current namespace.

bparees: i'm guessing the /etc/resolv.conf that gets mounted into the container includes a search domain that specifies the namespace.  it would be easy for you to confirm that.  (and easier than using the downward api).

but this still needs to be based on the service_name env variable, not hardcoded.

mnagy:
Fixed. We now use a separate environment variable for master service name, called '$MYSQL_MASTER_SERVICE_NAME'. PTAL

bparees:
lgtm.
[test]

bparees:
[test]

PR #87, opened 2015-08-26
Make assert_login_access fail when it should

### LINKED ISSUE #84: assert_login_access always pass when success=false
[`assert_login_access`](https://github.com/openshift/mysql/blob/04c1172948bf0a62a10b2bdbaaff75cea503f071/5.5/test/run#L158) always succeed.

That means that assertions like [`assert_login_access $USER "${PASS}_foo" false`](https://github.com/openshift/mysql/blob/04c1172948bf0a62a10b2bdbaaff75cea503f071/5.5/test/run#L261) never fail.

rhcarvalho:
Fixes #84 

@bparees @mfojtik PTAL

### DIFF
--- a/5.5/test/run
+++ b/5.5/test/run
@@ -152,11 +152,19 @@ function assert_login_access() {
   local PASS=$1 ; shift
   local success=$1 ; shift
 
-  if $success; then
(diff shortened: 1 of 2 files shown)


bparees:
lgtm.

PR #88, opened 2015-08-26
Reorder Dockerfile for better use of cache

rhcarvalho:
Every change to the scripts was causing yum install to run over and over.

@bparees PTAL

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -22,9 +22,6 @@ LABEL io.k8s.description="MySQL is a multi-user, multi-threaded SQL database ser
 
 EXPOSE 3306
 
-COPY run-*.sh /usr/local/bin/
-COPY contrib /var/lib/mysql/
-
 # This image must forever use UID 27 for mysql user so our volumes are
 # safe in the future. This should *never* change, the last test is there
 # to make sure of that.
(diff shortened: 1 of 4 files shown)


5.5/Dockerfile
```diff
@@ -36,6 +33,9 @@ RUN rpmkeys --import file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
     test "$(id mysql)" = "uid=27(mysql) gid=27(mysql) groups=27(mysql)" && \
     chmod o+w -R /var/lib/mysql
 
+COPY run-*.sh /usr/local/bin/
+COPY contrib /var/lib/mysql/
```
bparees: same comment as the other PR (but that chmod o+w should be a chmod a+rwx i think.  or at least a+rw)

rhcarvalho:
@bparees and this is the last change of this type for the time being. PTAL

bparees:
lgtm, will wait for test to finish.

mfojtik:
LGTM (do we need to change other dockerfiles?)

rhcarvalho:
@mfojtik I changed mongodb, postgresql, mysql. That might be it.

PR #92, opened 2015-09-09
WIP: Add support for atomic command without systemd unit file

hhorak:
This is reworked PR, similar to https://github.com/openshift/mysql/pull/83 but with these changes:
* no systemd unit file is created, because of https://github.com/projectatomic/atomic/issues/151
* atomic labels are much simpler
* options for docker container (MYSQL_USER and stuff) are not passed by `OPTx` variables, but instead all unknown options from `atomic install` are passed to `docker run` command
* `atomic run <container>` is used to start the created container
* `UNINSTALL` label doesn't have to be used, since only default action is enough

### DIFF
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -21,7 +21,8 @@ LABEL io.k8s.description="MySQL is a multi-user, multi-threaded SQL database ser
       io.openshift.tags="database,mysql,mysql55"
 
 # Support for atomic run/install/uninstall
-LABEL INSTALL="docker run --rm \${IMAGE} cat /usr/share/container-layer/mysql/atomic/install.sh | bash -s --"
+LABEL INSTALL="docker run --rm \"\${IMAGE}\" cat /usr/share/container-layer/mysql/atomic/install.sh | bash -s --" \
+      RUN="bash -c 'docker run --name \"\${NAME}\" \"$@\" \"\${IMAGE}\"' --"
 
 EXPOSE 3306
 
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -21,8 +21,7 @@ LABEL io.k8s.description="MySQL is a multi-user, multi-threaded SQL database ser
       io.openshift.tags="database,mysql,mysql55"
 
 # Support for atomic run/install/uninstall
-LABEL INSTALL="docker run --rm \${IMAGE} cat /usr/share/container-layer/mysql/atomic/install.sh | bash -s --" \
-      RUN="docker start \${NAME}"
+LABEL INSTALL="docker run --rm \${IMAGE} cat /usr/share/container-layer/mysql/atomic/install.sh | bash -s --"
 
 EXPOSE 3306
 
--- a/5.5/Dockerfile
+++ b/5.5/Dockerfile
@@ -20,6 +20,10 @@ LABEL io.k8s.description="MySQL is a multi-user, multi-threaded SQL database ser
       io.openshift.expose-services="3306:mysql" \
       io.openshift.tags="database,mysql,mysql55"
 
+# Support for atomic run/install/uninstall
+LABEL INSTALL="docker run --rm \${IMAGE} cat /usr/share/container-layer/mysql/atomic/install.sh | bash -s --" \
+      RUN="docker start \${NAME}"
+
 EXPOSE 3306
 
 # This image must forever use UID 27 for mysql user so our volumes are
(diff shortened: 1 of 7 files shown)


hhorak:
@mfojtik what about this variant, does it make better sense to you?

mfojtik:
@hhorak **much** better :-)

LGTM (wait for @bparees for merge)

mfojtik:
[test]

hhorak:
well, I'm still not sure about how the `atomic run` should behave, it's specified in the https://github.com/openshift/mysql/pull/83#issuecomment-138889244
Let's wait with merging until we're clear about it.

bparees:
@bparees waits.

rhatdan:
No reason for this:
```
+chcon -Rt svirt_sandbox_file_t "${DATADIR}"
```
If you are mounting with the :Z

rhatdan:
Not sure why you need this?
```
RUN chmod -R a+rwx /var/lib/mysql
```

Do users within the container write content in this directory?  They would not be allowed to write to the /var/lib/mysql/data directory since that is not loose like this?

hhorak:
> If you are mounting with the :Z

Thanks, will remove.

> Do users within the container write content in this directory? They would not be allowed to write to the /var/lib/mysql/data directory since that is not loose like this?

Yes, my.cnf is placed into `/var/lib/mysql/my.cnf` during container start. `chmod -R a+rwx` allows to run the container as any UID, which is one of the OpenShift's requirements.

rhatdan:
I have a real concern about doing this.  Since this directory becomes a world writable directory on disk, that would potentially allow users outside of the container to write there, which is frowned on by security people.  If this container were used by something other then openshift, it could be considered a security issue.  There has been some effort to allow you to specify a group to add a user to when running a container.  @mrunalp was working on it I believe.

Then you could chcon 770 /var/lib/mysql and add the user to the mysqld group on entry.

rhatdan:
I think the option is this, although not well defined.
```
  --group-add=[]                  Add additional groups to join
```

mrunalp:
Yes, that is the correct option. The groups could be specified by either group names or just GIDs. If names are specified, they are looked up within the container.

rhatdan:
@mrunalp The help and man page on this stinks.  :^(

mrunalp:
@rhatdan Sorry about that :( will send a PR to make them better.

hhorak:
> I think the option is this, although not well defined.
> 
>   --group-add=[]                  Add additional groups to join

Which command are we talking about?

rhatdan:
Something like the following.
docker run -user=1234 --group-add=mariadb mariadb

I beliew would run the conatainer as UID 1234 with 1234 being in the group mariadb.

hhorak:
@rhatdan I see, the `--group-add` is something to be added to `docker run`, right? It doesn't work yet.

Anyway, I miss how this helps us, since once we set `/var/lib/mysql` to 0770, users would specify `--group-add` every-time to be able to run the container. That doesn't seem to be very usable. 

> Since this directory becomes a world writable directory on disk, that would potentially allow users outside of the container to write there

Thinking a bit more about this, I don't get it. We change 0777 to `/var/lib/mysql` but what we volume mount is `/var/lib/mysql/data`, so this directory (the mounted one) has usual permissions on host. Maybe it's just too late to understand this :)

rhatdan:
> @rhatdan I see, the --group-add is something to be added to docker run, right? It doesn't work yet.
> 
Might be a new feature of docker that has not gotten out there yet.

> Anyway, I miss how this helps us, since once we set /var/lib/mysql to 0770, users would specify --group-add every-time to be able to run the container. That doesn't seem to be very usable.
>
Do you have to specify docker run --user?  If yes what is the problem in specifying --group-add?  I see this as something that could be wrapped or use atomic run.
  
>     Since this directory becomes a world writable directory on disk, that would potentially allow users outside of the container to write there
> 
> Thinking a bit more about this, I don't get it. We change 0777 to /var/lib/mysql but what we volume mount is /var/lib/mysql/data, so this directory (the mounted one) has usual permissions on host. Maybe it's just too late to understand this :)

/var/lib/mysql is a directory on disk that might be accessible to certain processes (ALthough you might be protected via mount namespaces)

hhorak:
I'll need to think more about it in the morning, but maybe we can set `USER 27:0` in the Dockerfile and having `/var/lib/mysql` with just `0770`. In case somebody runs the container with `-u 1000`, he still gets GID 0, so it should work.

hhorak:
@rhatdan https://github.com/openshift/mysql/pull/94 is attempt to make the image more safe. Feel free to comment there whether it is a good idea or not.

rhatdan:
That would actually make it less safe.  Basically would allow the process you are sticking into the container to have root group access within the container.  It would probably be ok to add the GID of mariadb though.

--user 1000:MARIADBGROUP

hhorak:
@rhatdan Well, yes, but that's how docker works, or more specifically how `--user` option of `docker` command works -- we can't prohibit users from using just `--user=1000`.

The default user for the image is not changed, it's still `mysql:mysql`.

The change in the PR above is basically just change from `drwxrwxrwx mysql mysql` to `drwxrwx--- mysql root` for `/var/lib/mysql` directory, which can't be less secure imho.

hhorak:
We can close this for now, until we know for sure what we expect from atomic support from non-system containers.

PR #94, opened 2015-09-11
Less loose permission bits for /var/lib/mysql inside container

hhorak:
As mentioned in https://github.com/openshift/mysql/pull/92#issuecomment-139270899,
using `0777` for `/var/lib/mysql` is not very safe. But we can allow to run the image
as arbitrary user even with `0770` bits, by setting `root` group as owner for `/var/lib/mysql`.

That works because when only specifying user (e.g. `-u 1000`), group 0 is used.


mfojtik:
[test]

PR #95, opened 2015-09-17
Port replication changes to 5.6

mnagy:


mnagy:
BTW, I also tested the replication with OS.

mfojtik:
@hhorak can you please review this?

mnagy:
@hhorak the variable was just renamed to MYSQL_RUNNING_AS_SLAVE to be consistent with 5.5 and more accurate. This variable is only used internally, so it shouldn't affect anything outside.

hhorak:
Sorry, seeing the response now.. Although I'd rather have the `MYSQL_DISABLE_CREATE_DB` option back and be part of the API.. But I'll create a separate PR for that, including documentation, so it is clear what it serves for. Otherwise it seems ok.
LGTM.

mnagy:
@hhorak Thanks a lot for the quick review! :) Merging.

PR #97, opened 2015-09-18
Make MYSQL_USER, MYSQL_PASSWORD and MYSQL_DATABASE optional

hhorak:
Implements https://github.com/openshift/mysql/issues/96

### DIFF
--- a/5.5/contrib/common.sh
+++ b/5.5/contrib/common.sh
@@ -26,6 +26,7 @@ function usage() {
   echo "  MYSQL_DATABASE (regex: '$mysql_identifier_regex')"
   echo "Or the following environment variable:"
   echo "  MYSQL_ROOT_PASSWORD (regex: '$mysql_password_regex')"
+  echo "Or both."
   echo "Optional Settings:"
   echo "  MYSQL_LOWER_CASE_TABLE_NAMES (default: 0)"
   echo "  MYSQL_MAX_CONNECTIONS (default: 151)"
--- a/5.5/contrib/common.sh
+++ b/5.5/contrib/common.sh
@@ -20,13 +20,13 @@ mysql_password_regex='^[a-zA-Z0-9_~!@#$%^&*()-=<>,.?;:|]+$'
 
 function usage() {
   [ $# == 2 ] && echo "error: $1"
-  echo "You must specify following environment variables:"
+  echo "You must either specify the following environment variables:"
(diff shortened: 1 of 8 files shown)


5.5/contrib/common.sh
```diff
@@ -20,13 +20,13 @@ mysql_password_regex='^[a-zA-Z0-9_~!@#$%^&*()-=<>,.?;:|]+$'
...
   [ $# == 2 ] && echo "error: $1"
-  echo "You must specify following environment variables:"
+  echo "You must either specify the following environment variables:"
   echo "  MYSQL_USER (regex: '$mysql_identifier_regex')"
   echo "  MYSQL_PASSWORD (regex: '$mysql_password_regex')"
   echo "  MYSQL_DATABASE (regex: '$mysql_identifier_regex')"
-  echo "Optional:"
+  echo "Or the following environment variable:"
   echo "  MYSQL_ROOT_PASSWORD (regex: '$mysql_password_regex')"
-  echo "Settings:"
```
bparees: "Or both"

bparees:
@mnagy please review.

5.5/contrib/common.sh
hhorak: Do you mean it this way?
```
...
Or the following environment variable:
  MYSQL_ROOT_PASSWORD (regex: '...')"
Or both.
Settings:
...
```

bparees: yes.      just trying to clarify to users that they can set both a ROOT password and a USER password, they aren't mutually exclusive, but at least one is required.

hhorak: Makes sense, fixed..

5.6/test/run
```diff
@@ -198,6 +197,7 @@ function try_image_invalid_combinations() {
 
 function run_container_creation_tests() {
   echo "  Testing image entrypoint usage"
+  assert_container_creation_fails "$@"
```
mnagy: We never call run_container_creation_tests with arguments, so `"$@"` here is redundant.

hhorak: Thanks, fixed..

mnagy:
This is looking good. I'm going to run some more tests later today/early tomorrow and will lgtm if they pass.

mnagy:
Tests have passed, LGTM.

PR #98, opened 2015-10-01
move newer image into centos/rhscl namespace

praiskup:
While we are on it, also unify the build scripts among images'
repositories:

- hack/ directory should be _just_ mirrored
- Makefile should be as declarative as possible
- build.sh should not rely on repository name

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,24 +1,7 @@
-SKIP_SQUASH?=0
-VERSIONS="5.5 5.6"
+# Variables are documented in hack/build.sh.
+BASE_IMAGE_NAME = mysql
+VERSIONS = 5.5 5.6
+OPENSHIFT_NAMESPACES = 5.5
 
-ifeq ($(TARGET),rhel7)
-	OS := rhel7
-else
-	OS := centos7
-endif
-
-ifeq ($(VERSION), 5.5)
-	VERSION := 5.5
-else ifeq ($(VERSION), 5.6)
-        VERSION := 5.6
-else
-	VERSION :=
-endif
-
-.PHONY: build
-build:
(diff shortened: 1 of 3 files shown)


bparees:
[test]

hack/common.mk
```diff
@@ -0,0 +1,27 @@
+SKIP_SQUASH?=0
+
+build = hack/build.sh
+
+NAMESPACE = normal
```
bparees: i don't see where this is used?

Makefile
```diff
@@ -1,24 +1,5 @@
-SKIP_SQUASH?=0
-VERSIONS="5.5 5.6"
+VERSIONS = 5.5 5.6
+OPENSHIFT_NAMESPACES = 5.5
```
bparees: please comment this.  i had to read through too much scripting to understand what it does.

hack/build.sh
```diff
@@ -1,19 +1,24 @@
 #!/bin/bash -e
 # This script is used to build, test and squash the OpenShift Docker images.
 #
-# $1 - Specifies distribution - "rhel7" or "centos7"
-# $2 - Specifies the image version - (must match with subdirectory in repo)
+# OS - Specifies distribution - "rhel7" or "centos7"
+# VERSION - Specifies the image version - (must match with subdirectory in repo)
```
bparees: you should add BASE_IMAGE and OPENSHIFT_NAMESPACES to the list of variables this script consumes.

praiskup:
PTAL

bparees:
lgtm, will wait for tests to pass.  @mfojtik @jhadvig any opinions on this refactor?

bparees:
[test]

mfojtik:
LGTM if we do the same for the rest of images.

PR #100, opened 2015-10-18
Refactoring and re-organizing non-RPM scripts

hhorak:
This is mostly a redesign of the scripts, while most changes were already consulted before and in https://github.com/openshift/mysql/pull/69. This PR is less creative about scripts structure and doesn't require from users to understand the structure.

All the changes are transparent for the container user, while they are just re-organizing the scripts. The reason for the changes is to allow to extend images easier for user in upper layer and also to make future maintenance simpler, since one complex bash script would be hard to change in another layer.

These are the changes made:
* available commands (that are expected to be used as docker run argument) are now called `run-mysqld`, `run-mysqld-master`, `run-mysqld-slave` (with backward compatible links `mysqld-master` and `mysqld-slave` and those binaries are located under `/usr/bin`, so users may start them on demand, e.g. when debugging the image.
* entrypoint script is now called `container-entrypoint`
* the entrypoint now serves just for changing environment, nothing more
* configuration files are located in the expected (`/etc/my.cnf`, `/etc/my.cnf.d`), so users may work with `/etc/my.cnf` and `/etc/my.cnf.d` directly as they are used to from RPMs.
* preparation of directories during build (fixing access) is now done by script `/usr/libexec/container-setup`, so the dockerfile is more clean
* images are now more easily extensible, by changing scripts that are located in the expected directory (e.g. `/usr/share/container-scripts/mysql/post-init.sh`)
* scripts don't include collection name hard-coded, but the collection is specified in the Dockerfile and thus most of the scripts may be common for more versions of mysql (which makes maintenance easier)


hhorak:
[test]

hhorak:
@mnagy any feedback? we need to know whether it is worth creating something similar for other database images..

mnagy:
Sorry, hadn't had time to look at this. I'll do it over the weekend.

hhorak:
ping

bparees:
@mnagy is on PTO now, i guess he didn't get to reviewing this before he left.  Given that we're trying to lock things down for 3.1, what's the urgency here?

hhorak:
@bparees If someone starts to depend on paths we're using insider container, then it may cause troubles in the future. But I think we can wait with this after 3.1, since depending on paths is not something we document, so users shouldn't take that as something that won't change.

mfojtik:
@bparees i'm going to review this and slap @mnagy when he will come back.

5.5/Dockerfile
```diff
@@ -27,30 +27,26 @@ EXPOSE 3306
...
 # the SCL for all scripts without need to do 'scl enable'.
-ENV BASH_ENV=/var/lib/mysql/scl_enable \
-    ENV=/var/lib/mysql/scl_enable \
-    PROMPT_COMMAND=". /var/lib/mysql/scl_enable"
+ENV BASH_ENV=/usr/share/container-scripts/mysql/scl_enable \
+    ENV=/usr/share/container-scripts/mysql/scl_enable \
+    PROMPT_COMMAND=". /usr/share/container-scripts/mysql/scl_enable" \
+    ENABLED_COLLECTIONS=mysql55
+
+ADD root /
```
mfojtik: @hhorak I do believe this change will go to all other images as well and this is just first PR to gather feedback.

5.5/Dockerfile.rhel7
```diff
@@ -33,30 +33,26 @@ EXPOSE 3306
...
 # When bash is started non-interactively, to run a shell script, for example it
 # looks for this variable and source the content of this file. This will enable
 # the SCL for all scripts without need to do 'scl enable'.
-ENV BASH_ENV=/var/lib/mysql/scl_enable \
-    ENV=/var/lib/mysql/scl_enable \
-    PROMPT_COMMAND=". /var/lib/mysql/scl_enable"
+ENV BASH_ENV=/usr/share/container-scripts/mysql/scl_enable \
+    ENV=/usr/share/container-scripts/mysql/scl_enable \
+    PROMPT_COMMAND=". /usr/share/container-scripts/mysql/scl_enable" \
+    ENABLED_COLLECTIONS=mysql55
```
mfojtik: isn't there an environment variable already with this information?

5.5/root/usr/bin/run-mysqld
```diff
@@ -0,0 +1,23 @@
...
+  start_local_mysql "$@"
+fi
+
+[ -f ${MYSQL_SCRIPTS_PATH}/passwd-change.sh ] && . ${MYSQL_SCRIPTS_PATH}/passwd-change.sh
+[ -f ${MYSQL_SCRIPTS_PATH}/post-init.sh ] && . ${MYSQL_SCRIPTS_PATH}/post-init.sh
+
+# Restart the MySQL server with public IP bindings
+mysqladmin $admin_flags flush-privileges shutdown
+unset_env_vars
+exec ${MYSQL_PREFIX}/libexec/mysqld --defaults-file=$MYSQL_DEFAULTS_FILE "$@" 2>&1
```
mfojtik: shouldn't this call the container entrypoint?

5.5/root/usr/bin/run-mysqld-master
```diff
@@ -0,0 +1,48 @@
...
+  GRANT SELECT ON replication.* TO '${MYSQL_MASTER_USER}'@'%' IDENTIFIED BY '${MYSQL_MASTER_PASSWORD}';
+  FLUSH PRIVILEGES;
+EOSQL
+
+[ -f ${MYSQL_SCRIPTS_PATH}/post-init.sh ] && . ${MYSQL_SCRIPTS_PATH}/post-init.sh
+
+# Restart the MySQL server with public IP bindings
+mysqladmin $admin_flags flush-privileges shutdown
+unset_env_vars
+exec ${MYSQL_PREFIX}/libexec/mysqld --defaults-file=$MYSQL_DEFAULTS_FILE "$@" 2>&1
```
mfojtik: same here

5.5/root/usr/share/container-scripts/mysql/common.sh
```diff
@@ -5,8 +5,14 @@
 # volume to override it.
 export MYSQL_DATADIR=/var/lib/mysql/data
 
+# Get prefix rather than hard-code it
+export MYSQL_PREFIX=$(which mysqld_safe|sed -e 's|/bin/mysqld_safe$||')
```
mfojtik: is there a reason why we can't hard code the prefix that is worth installing 'which' ? ;-)

5.5/root/usr/bin/run-mysqld-slave
```diff
@@ -2,23 +2,26 @@
 #
 # This is an entrypoint that runs the MySQL server in the 'slave' mode.
 #
-source ${HOME}/common.sh
+source /usr/share/container-scripts/mysql/common.sh
```
mfojtik: source ${MYSQL_SCRIPTS_PATH} ?

5.5/root/usr/share/container-scripts/mysql/common.sh
```diff
@@ -14,55 +20,13 @@ export MYSQL_FT_MAX_WORD_LEN=${MYSQL_FT_MAX_WORD_LEN:-20}
 export MYSQL_AIO=${MYSQL_AIO:-1}
 
 # Be paranoid and stricter than we should be.
-# https://dev.mysql.com/doc/refman/5.5/en/identifiers.html
+# https://dev.mysql.com/doc/refman/5.6/en/identifiers.html
```
mfojtik: we should not include links with versions as they can get old fast :-)

5.5/root/usr/share/container-scripts/mysql/scl_enable
```diff
@@ -1,3 +1,3 @@
 # This will make scl collection binaries work out of box.
 unset BASH_ENV PROMPT_COMMAND ENV
-source scl_source enable mysql55
+source scl_source enable ${ENABLED_COLLECTIONS}
```
mfojtik: i understand the variable above now, but don't understand the intention of it... do you want to allow people enabling additional collections when using mysql as a base image for something?

mfojtik:
@hhorak @bparees couple small nits, but in generally this looks great to me

5.5/Dockerfile
hhorak: Ideally yes.

5.5/Dockerfile.rhel7
hhorak: so far it has been only hardcoded in the `scl_enable` file.

5.5/root/usr/bin/run-mysqld
hhorak: I don't think so, the container-entrypoint script is already exec'ing this particular script.

5.5/root/usr/bin/run-mysqld-slave
hhorak: This variable is defined in this file. We can replace it if we define it directly in the Dockerfile, which seems like a good idae btw. Do you think so as well?

5.5/root/usr/share/container-scripts/mysql/common.sh
hhorak: This way we can share the scripts across the versions and SCL vs. non-SCL images. `which` RPM is only 77kB, does it seem to be worth the trouble of maintaining different versions of scripts to save that little space? From my PoV it does not.. But maybe we can avoid using which somehow.. just not having a quick idea right now.

5.5/root/usr/share/container-scripts/mysql/common.sh
hhorak: I agree and just found out the redirect magically works if we simply omit the version piece.. will do an update...

5.5/root/usr/share/container-scripts/mysql/scl_enable
hhorak: It's again more about being able to share the scripts between various versions and possibly even with mariadb..

5.5/root/usr/bin/run-mysqld-slave
```diff
@@ -2,23 +2,26 @@
...
 set -eu
 
-export MYSQL_RUNNING_AS_SLAVE=1
+# Just run normal server if the data directory is alreayd initialized
```
bparees: s/alreayd/already/

5.5/root/usr/share/container-scripts/mysql/common.sh
```diff
@@ -5,8 +5,14 @@
 # volume to override it.
 export MYSQL_DATADIR=/var/lib/mysql/data
 
+# Get prefix rather than hard-code it
+export MYSQL_PREFIX=$(which mysqld_safe|sed -e 's|/bin/mysqld_safe$||')
+
+# Path to MySQL scripts
+export MYSQL_SCRIPTS_PATH=/usr/share/container-scripts/mysql
```
bparees: as discussed above, defining this in the dockerfile seems reasonable.  perhaps give the variable a more generic name to make it easier to copy/paste it between dockerfiles for other products?

bparees:
[test]

5.5/root/usr/bin/run-mysqld-slave
hhorak: fixed

5.5/root/usr/share/container-scripts/mysql/common.sh
hhorak: Fixed now - defined in Dockerfile and renamed to `CONTAINER_SCRIPTS_PATH` so we can use the same variable in all containers.

hhorak:
@praiskup @omron93 What do you say about this PR? It's kind of a compromise between what we started with in summer and what we're having now. Since it seems it is acceptable for OpenShift, do you think something similar would be possible to do for mongodb and postgresql as well?

praiskup:
As PostgreSQL container works fine for OpenShift, and there are no other users
-- I would rather wait for what will be really needed.

I have no problem to submit PR to sync OpenShift-PostgreSQL image to contain
the same *important* concepts that will come up from this discussion (if any
user can benefit from that) once this PR is done.

What I like on this PR:
- entrypoint does not take '$1' == 'postgres' as special (unexpectable) action
- entrypoint's script name is defined

But:
- entrypoint is now redundant (at least in this PR)
- I can't put something into /usr/share/container-scripts/mysql during
  'docker build' time and then allow users to bind-mount script directory
  into /usr/shar/container/scripts/mysql ..  but this is not needed (and its
  not expected to be) by OpenShfit anyway.

hhorak:
@praiskup thanks for the feedback.

Having working image is the main goal, yes. At the same time, these images will be taken as example by others and we will therefore define the best practices by preparing the images the best we can (except writing-up the best practices as well). That said, having the images following similar style is actually important feature. We can talk about it more f2f.

> But:
> 
>     entrypoint is now redundant (at least in this PR)

The entrypoint is still necessary to allow doing something like this: `docker run IMGNAME psql` -- without entrypoint the command `psql` is not found.

>     I can't put something into /usr/share/container-scripts/mysql during 'docker build' time and then allow users to bind-mount script directory into /usr/shar/container/scripts/mysql .. but this is not needed (and its not expected to be) by OpenShfit anyway.

Correct, although the bind-mount use case might work with separate files, but you're right this is not something we support now.

mfojtik:
@hhorak i'm OK to merge this, can you please squash?

hhorak:
@mfojtik squashed. Thanks!

PR #101, opened 2015-10-19
Use explicit rhel7 version

hhorak:
Using `rhel7` only doesn't have to correspond with what we expect (latest rhel7), so we need to be more explicit here.

### DIFF
--- a/5.5/Dockerfile.rhel7
+++ b/5.5/Dockerfile.rhel7
@@ -1,4 +1,4 @@
-FROM rhel7
+FROM rhel7.1
 
 # MySQL image for OpenShift.
 #
--- a/5.6/Dockerfile.rhel7
+++ b/5.6/Dockerfile.rhel7
@@ -1,4 +1,4 @@
-FROM rhel7
+FROM rhel7.1
 
 # MySQL image for OpenShift.
 #


bparees:
this is failing because rhel7 is actually the tag we use on our AMI that does the docker build, it's an internal version of the rhel7 image that has subscription certs added so the yum install commands will work.

we'll need to update our AMI process to create a rhel7.1 tag if you want this to succeed in our build system, but even that you're not actually going to be getting something that's necessarily based on rhel7.1.

hhorak:
So, from my PoV, correct `FROM rhel7` is important for users rebuilding the image themselves and they should simply use `rhel7.1` provided by RH. As for the CI, I think we should be fine with even fake rhel7.1, if that's something you propose..

bparees:
btw is there a reason we don't want to just pick up whatever rhel7 points to?  (i assume someday rhel7 will pick up rhel7.2, for example).  is that a bad thing?

hhorak:
The reason is that it just doesn't work this way :) I was told that `rhel7:latest` may point to `rhel7.1` or `rhel7.2`, depending on build order, one of them every-time.

bparees:
that's terrifying :)

bparees:
ami has been updated to have a rhel7.1 image.  [test]

PR #104, opened 2015-11-02
workaround for squash rm issues

bparees:


bparees:
@mfojtik ptal

bparees:
@hhorak fyi

mfojtik:
@bparees LGTM but still interested to hear from Marek what is the root cause of this

mfojtik:
@bparees need extra step, this one breaks as well

bparees:
@mfojtik thanks. updated.

PR #109, opened 2015-11-10
Handle MYSQL_USER=root case

mnagy:
The MYSQL_DATABASE is now optional, if the root is specified. No
documentation in usage was added, since this is a real edge case and we
do not want to make the usage needlessly hard to read.

@hhorak fyi

This fixes https://bugzilla.redhat.com/show_bug.cgi?id=1262861

### DIFF
--- a/5.5/root/usr/share/container-scripts/mysql/validate-variables.sh
+++ b/5.5/root/usr/share/container-scripts/mysql/validate-variables.sh
@@ -18,12 +18,10 @@ function usage() {
 
 function validate_variables() {
   # Check basic sanity of specified variables
-  if [[ -v MYSQL_USER && -v MYSQL_PASSWORD && -v MYSQL_DATABASE ]]; then
+  if [[ -v MYSQL_USER && -v MYSQL_PASSWORD ]]; then
     [[ "$MYSQL_USER"     =~ $mysql_identifier_regex ]] || usage "Invalid MySQL username"
     [ ${#MYSQL_USER} -le 16 ] || usage "MySQL username too long (maximum 16 characters)"
     [[ "$MYSQL_PASSWORD" =~ $mysql_password_regex   ]] || usage "Invalid password"
-    [[ "$MYSQL_DATABASE" =~ $mysql_identifier_regex ]] || usage "Invalid database name"
-    [ ${#MYSQL_DATABASE} -le 64 ] || usage "Database name too long (maximum 64 characters)"
     user_specified=1
   fi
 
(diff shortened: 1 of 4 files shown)


hhorak:
Makes sense, LGTM.

5.5/root/usr/share/container-scripts/mysql/validate-variables.sh
```diff
@@ -32,11 +30,34 @@ function validate_variables() {
     root_specified=1
   fi
 
+  # If MYSQL_USER == "root", we have a special case
+  if [[ "${user_specified:-0}" == "1" && "$MYSQL_USER" == "root" ]]; then
+    if [[ "${root_specified:-0}" == "1" ]]; then
+      usage "You cannot set MYSQL_USER to 'root' and use MYSQL_PASSWORD and MYSQL_ROOT_PASSWORD at the same time"
```
bparees: shouldn't this just say "you can't set mysql_user to 'root' and use mysql_root_password at the same time?

5.5/root/usr/share/container-scripts/mysql/validate-variables.sh
```diff
@@ -32,11 +30,34 @@ function validate_variables() {
...
+    user_specified=0
+    root_specified=1
+  fi
+
   # Either combination of user/pass/db or root password is ok
   if [[ "${user_specified:-0}" == "0" && "${root_specified:-0}" == "0" ]]; then
     usage
   fi
 
+  # If the root user is not specified, database name is required
```
bparees: why are we allowing you to specify a root password with no database name?

5.5/root/usr/share/container-scripts/mysql/validate-variables.sh
bparees: nm, I see what you're saying/doing here... if mysql_user==root, then mysql_password and mysql_root_password become equivalent, so specifying both is an error.

reword the message to:
usage "When setting MYSQL_USER to 'root' you can only set either MYSQL_PASSWORD or MYSQL_ROOT_PASSWORD"

though I still think all this behavior is confusing and would have gone for just reporting an error if they set MYSQL_USER to root and say "don't do that, to configure the root user just use MYSQL_ROOT_PASSWORD and do not specify MYSQL_USERNAME"

hhorak: This solution would make sense to me as well, although I'm not entirely convinced whether one of the solution is better than the other; this one would be simpler to implement at least.

5.5/root/usr/share/container-scripts/mysql/validate-variables.sh
mnagy: Because I've realized that we do not really need the database. You actually only need it in a replication setup, where you need to make sure it is listed in the configuration files so MySQL replicates it.I've gone through the code and the change should work, we don't use the variable anywhere else.

5.5/root/usr/share/container-scripts/mysql/validate-variables.sh
bparees: i guess the existing implementation has the benefit of not failing/forcing a user to see the error and make a change.

bparees:
lgtm.

PR #114, opened 2015-12-01
Change RH required labels based on updated spec

hhorak:
For full spec, see https://github.com/projectatomic/ContainerApplicationGenericLabels/blob/master/vendor/redhat/labels.md

### DIFF
--- a/5.5/Dockerfile.rhel7
+++ b/5.5/Dockerfile.rhel7
@@ -13,17 +13,18 @@ FROM rhel7.2
 ENV MYSQL_VERSION=5.5 \
     HOME=/var/lib/mysql
 
-LABEL io.k8s.description="MySQL is a multi-user, multi-threaded SQL database server" \
+LABEL summary="MySQL is a multi-user, multi-threaded SQL database server" \
+      io.k8s.description="MySQL is a multi-user, multi-threaded SQL database server" \
       io.k8s.display-name="MySQL 5.5" \
       io.openshift.expose-services="3306:mysql" \
       io.openshift.tags="database,mysql,mysql55"
 
 # Labels consumed by Red Hat build service
-LABEL BZComponent="mysql55-docker" \
-      Name="openshift3/mysql-55-rhel7" \
-      Version="5.5" \
-      Release="1" \
-      Architecture="x86_64"
+LABEL com.redhat.component="mysql55-docker" \
+      name="openshift3/mysql-55-rhel7" \
+      version="5.5" \
+      release="1" \
+      architecture="x86_64"
 
 EXPOSE 3306
 
--- a/5.6/Dockerfile.rhel7
+++ b/5.6/Dockerfile.rhel7
@@ -13,17 +13,18 @@ FROM rhel7.2
 ENV MYSQL_VERSION=5.6 \
     HOME=/var/lib/mysql
 
-LABEL io.k8s.description="MySQL is a multi-user, multi-threaded SQL database server" \
+LABEL summary="MySQL is a multi-user, multi-threaded SQL database server" \
+      io.k8s.description="MySQL is a multi-user, multi-threaded SQL database server" \
       io.k8s.display-name="MySQL 5.6" \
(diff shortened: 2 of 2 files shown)


5.5/Dockerfile.rhel7
```diff
@@ -13,17 +13,18 @@ FROM rhel7.2
...
 # Labels consumed by Red Hat build service
-LABEL BZComponent="mysql55-docker" \
-      Name="openshift3/mysql-55-rhel7" \
-      Version="5.5" \
-      Release="1" \
-      Architecture="x86_64"
+LABEL com.redhat.component="mysql55-docker" \
+      name="openshift3/mysql-55-rhel7" \
+      version="5.5" \
+      release="1" \
```
mnagy: The proposal says that release label should be "The build of the image, `$IMGBUILD`". Shouldn't this be managed from outside? Will we have to bump this with every new release?

hhorak: This is one downstream change that we must do when building the image. It is even probable that we take upstream code at one point, then need to do more builds (from whatever reason) and so we need bump the release several times without any change in the upstream. So, I don't think this needs to be handled here, in upstream. But on the other hand I'd rather keep it here, even if it is not updated.

5.5/Dockerfile.rhel7
```diff
@@ -13,17 +13,18 @@ FROM rhel7.2
...
       io.openshift.expose-services="3306:mysql" \
       io.openshift.tags="database,mysql,mysql55"
 
 # Labels consumed by Red Hat build service
-LABEL BZComponent="mysql55-docker" \
-      Name="openshift3/mysql-55-rhel7" \
-      Version="5.5" \
-      Release="1" \
-      Architecture="x86_64"
+LABEL com.redhat.component="mysql55-docker" \
```
mnagy: Our component is "Image/Cartridge" with product "Atomic Enterprise Platform and OpenShift Enterprise". I don't see how to enter the "product". @hhorak, you're more bugzilla-savy than I, am I missing something here?

mnagy:
I found three missing labels: authoritative-source, vendor and distribution-scope.

5.5/Dockerfile.rhel7
mnagy: So you propose to have this here, just set to 1? And it'll then be updated downstream when building/releasing? If so, I'd rather use something that is not a valid release value, but still be admissible. How about setting it to 0?

hhorak: IMHO, users should be able to build the image as it is, so 1 seems better to me here (even though I'm not sure whether 0 would prevent anybody from building it).

5.5/Dockerfile.rhel7
mnagy: @bparees ?

bparees: i don't really know the exact purpose of these labels, but it was mysql55-docker before, so if it was ok then, i'm ok leaving it.

bparees:
@hhorak your other PRs didn't have the labels @mnagy is asking about either, so i'm going to merge this for consistency, but i am curious if we're going to need to add those other ones too?

hhorak:
`distribution-scope` is optional and to be honest we haven't talked about redistribution policy, so I don't even know what to set for this. Anyway, this one should imho be set by OSBS for all images and this label shouldn't be listed in the upstream Dockerfile.

For `authoritative-source` and `vendor`, these are already added by OSBS (well, older variant `Authoritative_Registry` is added ATM, but this will change), so they are available in the RHEL-based images. `Vendor` is also inherited from base image (`CentOS` value), but `authoritative-source` is not. 

However, if `authoritative-source` is only meant to be able to reconstruct the `docker pull ...` command, if `authoritative-source` is empty, then it results in correct `docker pull ...` command in case of CentOS-based images. So, `authoritative-source` shouldn't be needed there. I've reported that as https://github.com/projectatomic/ContainerApplicationGenericLabels/issues/65.

PR #116, opened 2015-12-16
Fix usage error output

mnagy:
The usage error was not printed out because we were checking for wrong
number of arguments.


mfojtik:
LGTM

PR #124, opened 2016-03-02
Expand container-entrypoint args as separate words

rhcarvalho:
Equivalent to https://github.com/openshift/sti-base/pull/80


rhcarvalho:
@mfojtik @bparees PTAL

PR #127, opened 2016-03-16
Grant the grant privilege to the root user

### LINKED ISSUE #125: Create multiple databases
When using root access I am not able to create multiple databases inside a single container. I get grant privileges error.

```
mysql> CREATE DATABASE demoDB;
Query OK, 1 row affected (0.00 sec)
mysql> GRANT ALL PRIVILEGES ON demoDB.* TO 'demo'@'%' IDENTIFIED BY 'demo';
ERROR 1044 (42000): Access denied for user 'root'@'%' to database 'demoDB'
```

To resolve this I have to:
```
mysql> UPDATE mysql.user SET Grant_priv='Y', Super_priv='Y' WHERE User='root';
mysql> FLUSH PRIVILEGES;
mysql> GRANT ALL ON *.* TO 'root'@'localhost';
Query OK, 0 rows affected (0.00 sec)
mysql> GRANT ALL PRIVILEGES ON demoDB.* TO 'demo'@'%' IDENTIFIED BY 'demo';
Query OK, 0 rows affected (0.00 sec)
```
By default, root user does not have grant and super privileges.

Any help is appreciated. I need to make sure the mysql 5.6 images works out of the box for me otherwise I will have to rebuild a new image.

mnagy:
Should enable root to set privileges for databases to other users. Root already has this privilege, but only by default when connecting from localhost. After this fix, it should work even when connecting remotely.

Fixes #125.

### DIFF
--- a/5.5/root/usr/share/container-scripts/mysql/common.sh
+++ b/5.5/root/usr/share/container-scripts/mysql/common.sh
@@ -115,7 +115,7 @@ EOSQL
   if [ -v MYSQL_ROOT_PASSWORD ]; then
     log_info "Setting password for MySQL root user ..."
 mysql $mysql_flags <<EOSQL
-    GRANT ALL PRIVILEGES ON *.* TO 'root'@'%' IDENTIFIED BY '${MYSQL_ROOT_PASSWORD}';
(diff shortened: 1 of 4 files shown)


bparees:
lgtm

PR #129, opened 2016-04-13
README.md: add example for pulling 5.6

php-coder:
Emphasize that images for 5.5 and 5.6 reside in a different namespaces.

PTAL @bparees @mfojtik


mfojtik:
LGTM

PR #130, opened 2016-04-15
[WIP] Porting MySQL tuning from V2

php-coder:
This PR ports MySQL configuration and tuning from [V2 cartridge](https://github.com/openshift/origin-server/blob/73197c277b9ce9044be67a4e23b9951d120bc663/cartridges/openshift-origin-cartridge-mysql/conf/my.cnf.erb).

It also adds new parameters that can be configured:
- `MYSQL_MAX_ALLOWED_PACKET`
- `MYSQL_TABLE_OPEN_CACHE`
- `MYSQL_SORT_BUFFER_SIZE`
- `MYSQL_KEY_BUFFER_SIZE`
- `MYSQL_READ_BUFFER_SIZE`
- `MYSQL_INNODB_BUFFER_POOL_SIZE`
- `MYSQL_INNODB_LOG_FILE_SIZE`
- `MYSQL_INNODB_LOG_BUFFER_SIZE`

When container has limited amount of memory a couple of parameters (`MYSQL_KEY_BUFFER_SIZE` and `MYSQL_INNODB_*`) calculated dynamically.

I changed a few default values:
- `table_open_cache` (4 -> 400, because 4 is too small value and I decided to use default value from MySQL itself)
- `innodb_log_file_size` (16M -> 8M, because there is a comment that suggest to set it to 25 % of buffer pool size that has default value 32M)
-  `innodb_log_buffer_size` (16M -> 8M, because there is a comment that suggests to set it to 25 % of buffer pool size that has default value 32M)

Questions for discuss:
- I didn't port parameters for `default-storage-engine` and `default-time-zone` because they require special bash-coding. Sshould I port them too?
- All configuration in `/etc/my.cnf.d/tuning.cnf` -- is it normal file name? Maybe it's better to rename it or even merge with already existing configs?

PTAL @bparees @rhcarvalho @mnagy @mfojtik 

Related issues:
- https://trello.com/c/JH769zUA/893-3-update-mysql-image-to-autoconfigure-based-on-available-memory-evg
- https://trello.com/c/p3y4Glt4/902-add-max-allowed-packet-as-an-environment-variable-to-the-mysql-images
- https://bugzilla.redhat.com/show_bug.cgi?id=1326523

Need to be done:
- [x] write tests
- [x] `README.md`: document auto-tuning
- [ ] update openshift-docs
- [ ] apply changes to 5.6 (after getting feedback here)
- [ ] ensure that MySQL replication works

### DIFF
--- a/5.5/root/usr/bin/cgroup-limits
+++ b/5.5/root/usr/bin/cgroup-limits
@@ -1,0 +1,92 @@
+#!/usr/bin/python
+
+"""
+Script for parsing cgroup information
+
+This script will read some limits from the cgroup system and parse
+them, printing out "VARIABLE=VALUE" on each line for every limit that is
+successfully read. Output of this script can be directly fed into
+bash's export command. Recommended usage from a bash script:
+
+    set -o errexit
+    export_vars=$(cgroup-limits) ; export $export_vars
+
+Variables currently supported:
+    MAX_MEMORY_LIMIT_IN_BYTES
+        Maximum possible limit MEMORY_LIMIT_IN_BYTES can have. This is
+        currently constant value of 9223372036854775807.
+    MEMORY_LIMIT_IN_BYTES
+        Maximum amount of user memory in bytes. If this value is set
+        to the same value as MAX_MEMORY_LIMIT_IN_BYTES, it means that
+        there is no limit set. The value is taken from
+        /sys/fs/cgroup/memory/memory.limit_in_bytes
+    NUMBER_OF_CORES
+        Number of detected CPU cores that can be used. This value is
+        calculated from /sys/fs/cgroup/cpuset/cpuset.cpus
+    NO_MEMORY_LIMIT
+        Set to "true" if MEMORY_LIMIT_IN_BYTES is so high that the caller
+        can act as if no memory limit was set. Undefined otherwise.
+"""
+
+from __future__ import print_function
+import sys
+
+
+def _read_file(path):
+    try:
+        with open(path, 'r') as f:
+            return f.read().strip()
+    except IOError:
+        return None
+
+
+def get_memory_limit():
+    """
+    Read memory limit, in bytes.
+    """
+
+    limit = _read_file('/sys/fs/cgroup/memory/memory.limit_in_bytes')
+    if limit is None or not limit.isdigit():
+        print("Warning: Can't detect memory limit from cgroups",
+              file=sys.stderr)
+        return None
+    return int(limit)
+
+
+def get_number_of_cores():
+    """
+    Read number of CPU cores.
+    """
+
+    core_count = 0
+
+    line = _read_file('/sys/fs/cgroup/cpuset/cpuset.cpus')
+    if line is None:
+        print("Warning: Can't detect number of CPU cores from cgroups",
+              file=sys.stderr)
+        return None
+
+    for group in line.split(','):
+        core_ids = list(map(int, group.split('-')))
+        if len(core_ids) == 2:
+            core_count += core_ids[1] - core_ids[0] + 1
+        else:
+            core_count += 1
+
+    return core_count
+
+
+if __name__ == "__main__":
+    env_vars = {
+        "MAX_MEMORY_LIMIT_IN_BYTES": 9223372036854775807,
+        "MEMORY_LIMIT_IN_BYTES": get_memory_limit(),
+        "NUMBER_OF_CORES": get_number_of_cores()
+    }
+
+    env_vars = {k: v for k, v in env_vars.items() if v is not None}
+
+    if env_vars.get("MEMORY_LIMIT_IN_BYTES", 0) >= 92233720368547:
+        env_vars["NO_MEMORY_LIMIT"] = "true"
+
+    for key, value in env_vars.items():
+        print("{0}={1}".format(key, value))
--- a/5.5/root/usr/bin/run-mysqld
+++ b/5.5/root/usr/bin/run-mysqld
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+export_vars=$(cgroup-limits); export $export_vars
 source ${CONTAINER_SCRIPTS_PATH}/common.sh
 set -eu
 
@@ -11,6 +12,7 @@ log_volume_info $MYSQL_DATADIR
 log_info 'Processing MySQL configuration files ...'
 envsubst < ${CONTAINER_SCRIPTS_PATH}/my-base.cnf.template > /etc/my.cnf.d/base.cnf
 envsubst < ${CONTAINER_SCRIPTS_PATH}/my-paas.cnf.template > /etc/my.cnf.d/paas.cnf
+envsubst < ${CONTAINER_SCRIPTS_PATH}/my-tuning.cnf.template > /etc/my.cnf.d/tuning.cnf
 
 if [ ! -d "$MYSQL_DATADIR/mysql" ]; then
   initialize_database "$@"
--- a/5.5/root/usr/share/container-scripts/mysql/common.sh
+++ b/5.5/root/usr/share/container-scripts/mysql/common.sh
@@ -14,6 +14,28 @@ export MYSQL_MAX_CONNECTIONS=${MYSQL_MAX_CONNECTIONS:-151}
 export MYSQL_FT_MIN_WORD_LEN=${MYSQL_FT_MIN_WORD_LEN:-4}
 export MYSQL_FT_MAX_WORD_LEN=${MYSQL_FT_MAX_WORD_LEN:-20}
 export MYSQL_AIO=${MYSQL_AIO:-1}
+export MYSQL_MAX_ALLOWED_PACKET=${MYSQL_MAX_ALLOWED_PACKET:-200M}
+export MYSQL_TABLE_OPEN_CACHE=${MYSQL_TABLE_OPEN_CACHE:-400}
+export MYSQL_SORT_BUFFER_SIZE=${MYSQL_SORT_BUFFER_SIZE:-256K}
+
+if [ -n "${NO_MEMORY_LIMIT:-}" -o -z "${MEMORY_LIMIT_IN_BYTES:-}" ]; then
+  key_buffer_size='32M'
+  read_buffer_size='8M'
+  innodb_buffer_pool_size='32M'
+  innodb_log_file_size='8M'
+  innodb_log_buffer_size='8M'
+else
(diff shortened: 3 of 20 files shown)


5.5/root/usr/share/container-scripts/mysql/my-paas.cnf.template
```diff
@@ -9,12 +9,15 @@ lower_case_table_names = ${MYSQL_LOWER_CASE_TABLE_NAMES}
 # The maximum permitted number of simultaneous client connections. Default: 151
 max_connections = ${MYSQL_MAX_CONNECTIONS}
 
-# The minimum length of the word to be included in a FULLTEXT index. Default: 4
+# The minimum/maximum lengths of the word to be included in a FULLTEXT index. Default: 4/20
```
bparees: not sure the point of this change?  not that it matters much either way.

5.5/root/usr/share/container-scripts/mysql/my-paas.cnf.template
```diff
@@ -9,12 +9,15 @@ lower_case_table_names = ${MYSQL_LOWER_CASE_TABLE_NAMES}
...
 ft_max_word_len = ${MYSQL_FT_MAX_WORD_LEN}
 
 # In case the native AIO is broken. Default: 1
 # See http://help.directadmin.com/item.php?id=529
 innodb_use_native_aio = ${MYSQL_AIO}
+
+[myisamchk]
+# The minimum/maximum lengths of the word to be included in a FULLTEXT index. Default: 4/20
+ft_min_word_len = ${MYSQL_FT_MIN_WORD_LEN}
+ft_max_word_len = ${MYSQL_FT_MAX_WORD_LEN}
```
bparees: what's the story behind supplying these twice?  are they configuring two different things in mysql?  do the values have to match in both places?

5.5/root/usr/share/container-scripts/mysql/my-tuning.cnf.template
```diff
@@ -0,0 +1,28 @@
...
+read_buffer_size = ${MYSQL_READ_BUFFER_SIZE}
+read_rnd_buffer_size = 256K
+net_buffer_length = 2K
+thread_stack = 256K
+myisam_sort_buffer_size = 2M
+
+# It is recommended that innodb_buffer_pool_size is configured to 50 to 75 percent of system memory.
+innodb_buffer_pool_size = ${MYSQL_INNODB_BUFFER_POOL_SIZE}
+innodb_additional_mem_pool_size = 2M
+# Set .._log_file_size to 25 % of buffer pool size
```
bparees: this says 25% but you're actually setting it to 15% by default....

5.5/root/usr/share/container-scripts/mysql/my-paas.cnf.template
php-coder: Yes, I've just put together similar parameters. I can remove it if you want.

5.5/root/usr/share/container-scripts/mysql/my-paas.cnf.template
php-coder: Yes, they should match together, as far I understand. I found them while merging our and V2 configurations. Quote [from documentation](http://dev.mysql.com/doc/refman/5.7/en/myisamchk-general-options.html):
>To ensure that myisamchk and the server use the same values for full-text parameters, you can place each one in both the `[mysqld]` and `[myisamchk]` sections of an option file:
>```
>[mysqld]
>ft_min_word_len=3
>
>[myisamchk]
>ft_min_word_len=3```

5.5/root/usr/share/container-scripts/mysql/my-tuning.cnf.template
php-coder: Yes, it should be 25% of the buffer pool size that should be 50% of the memory. V2 rounded up 12.5% to 15%.

5.5/root/usr/share/container-scripts/mysql/my-paas.cnf.template
bparees: it's fine

5.5/root/usr/share/container-scripts/mysql/my-paas.cnf.template
bparees: maybe put a comment in the file then, explaining why they appear twice and that they should match, for future visitors to this file.

5.5/root/usr/share/container-scripts/mysql/my-tuning.cnf.template
bparees: ah... "of the buffer pool".  I should read things more fully :)

bparees:
> I didn't port parameters for default-storage-engine and default-time-zone because they require special bash-coding. Should I port them too?

let's defer it unless/until we get a request.

>  All configuration in /etc/my.cnf.d/tuning.cnf -- is it normal file name? Maybe it's better to rename it or even merge with already existing configs?

Having it as a separate file is ok with me and the name is fine.

5.5/root/usr/share/container-scripts/mysql/common.sh
```diff
@@ -14,6 +14,28 @@ export MYSQL_MAX_CONNECTIONS=${MYSQL_MAX_CONNECTIONS:-151}
...
+  read_buffer_size='8M'
+  innodb_buffer_pool_size='32M'
+  innodb_log_file_size='8M'
+  innodb_log_buffer_size='8M'
+else
+  key_buffer_size="$(python -c "print(int((${MEMORY_LIMIT_IN_BYTES}/(1024*1024))*0.1))")M"
+  read_buffer_size="$(python -c "print(int((${MEMORY_LIMIT_IN_BYTES}/(1024*1024))*0.05))")M"
+  innodb_buffer_pool_size="$(python -c "print(int((${MEMORY_LIMIT_IN_BYTES}/(1024*1024))*0.5))")M"
+  innodb_log_file_size="$(python -c "print(int((${MEMORY_LIMIT_IN_BYTES}/(1024*1024))*0.15))")M"
+  innodb_log_buffer_size="$(python -c "print(int((${MEMORY_LIMIT_IN_BYTES}/(1024*1024))*0.15))")M"
```
mnagy: Can't we calculate this using plain bash? I can get the same effect simpler:
```
$ echo "$(python -c "print(int((${MEMORY_LIMIT_IN_BYTES}/(1024*1024))*0.05))")M"
439804651110M
$ echo "$((MEMORY_LIMIT_IN_BYTES / 1024 / 1024 / 20))M"
439804651110M
```

php-coder: Probably, we can. I've copied this from @mfojtik's PR for PostgreSQL.

@mfojtik Why did you use Python here. I want to have similar code in different repos and if you update your PR I update mine.

mnagy:
You'll also need variable validation, especially if you want to use suffixes like K or M.

php-coder:
>You'll also need variable validation, especially if you want to use suffixes like K or M.

I'm not sure that I should add it. Currently we're validating only required variables.

mnagy:
That means we should ladd validation for those as well :)

php-coder:
>>>You'll also need variable validation, especially if you want to use suffixes like K or M.
>>I'm not sure that I should add it. Currently we're validating only required variables.
>That means we should ladd validation for those as well :)

For the report: we decided to not doing it on today's scrum meeting.

php-coder:
@bparees I've added a more doc and tests. PTAL. If you OK with it then I will apply this changes for 5.6 and prepare for merging.

5.5/root/usr/bin/cgroup-limits
```diff
@@ -0,0 +1,92 @@
...
+            core_count += core_ids[1] - core_ids[0] + 1
+        else:
+            core_count += 1
+
+    return core_count
+
+
+if __name__ == "__main__":
+    env_vars = {
+        "MAX_MEMORY_LIMIT_IN_BYTES": 9223372036854775807,
```
bparees: this isn't actually used right?

bparees: nm i guess it's provided to the caller's environment.

php-coder: Yes, it's unused. BTW, `NUMBER_OF_CORES` unused too. But they're here only to make it easier to sync this script with his upstream version.

bparees:
lgtm (ready for porting to 5.6).  nice job @php-coder!

5.5/root/usr/bin/cgroup-limits
rhcarvalho: is there any difference from the one in sti-base?

5.5/root/usr/bin/run-mysqld
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+export_vars=$(cgroup-limits); export $export_vars
```
rhcarvalho: Do we need to have `export_vars`? Why not `export $(cgroup-limits)`?

5.5/root/usr/share/container-scripts/mysql/common.sh
rhcarvalho: What's the conclusion on this?

5.5/root/usr/bin/cgroup-limits
php-coder: No, I've just copy it.

5.5/root/usr/bin/run-mysqld
php-coder: I don't know, I've used the same code as @mfojtik. My idea was to do the same things similar in a different places.

5.5/root/usr/share/container-scripts/mysql/common.sh
php-coder: Leave it as is unless @mfojtik will change it in his PR for PostgreSQL.

php-coder:
@bparees It's ready for merge.

5.5/root/usr/bin/run-mysqld
mfojtik: @mnagy can explainthe export vars

mnagy: It's there in the usage of the script :) You just forgot to use set -o errexit. The reason it is two separate commands is so that if cgroup-limits fails for some reason, the whole script fails with it. What I do not understand is why this was called before the source and set lines bellow.

PR #132, opened 2016-04-22
5.6/examples/replica/README.md: correct image name

php-coder:
PTAL @mnagy @bparees


bparees:
Doh. Good catch, lgtm.

PR #134, opened 2016-04-28
5.x/test/run: remove -x option

php-coder:
Disable `-x` flag and also filter out some diagnostic messages to make output clearly.

PTAL @bparees

### DIFF
--- a/5.5/test/run
+++ b/5.5/test/run
@@ -6,7 +6,8 @@
 # The image has to be available before this script is executed.
 #
 
-set -exo nounset
+set -o errexit
+set -o nounset
 shopt -s nullglob
 
 IMAGE_NAME=${IMAGE_NAME-openshift/mysql-55-centos7-candidate}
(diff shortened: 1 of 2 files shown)


5.5/test/run
```diff
@@ -108,15 +110,15 @@ function run_change_password_test() {
...
+  docker stop $(get_cid testpass1) >/dev/null
 
   # Create second container with changed password
   create_container "testpass2" -e MYSQL_USER=user -e MYSQL_PASSWORD=bar \
     -e MYSQL_DATABASE=db -v ${tmpdir}:/var/lib/mysql/data:Z
   test_connection testpass2 user bar
 
   # The old password should not work anymore
-  if mysql_cmd "$(get_container_ip testpass2)" user foo -e 'SELECT 1;'; then
+  if mysql_cmd "$(get_container_ip testpass2)" user foo -e 'SELECT 1;' 2>/dev/null; then
```
mfojtik: why we throwing errors ?

php-coder: To make output clean. For example, it was like this before:
```
ERROR 2003 (HY000): Can't connect to MySQL server on '172.17.0.7' (111)
ERROR 2003 (HY000): Can't connect to MySQL server on '172.17.0.7' (111)
172.17.0.8 successfully registered as SLAVE for 172.17.0.7
```
And now there is no such errors that actually isn't errors.

Do you think that we shouldn't hide them?

mfojtik: if we fail to connect for whatever reason, the reason will be hidden and it will be pain to debug it ;-)

php-coder:
@mfojtik Ok, I've removed `2>/dev/null`

bparees:
lgtm

PR #142, opened 2016-05-19
Fix image name for 5.6

mnagy:
This is just a tiny deduplication change, will be helpful later.

### DIFF
--- a/5.6/Dockerfile.rhel7
+++ b/5.6/Dockerfile.rhel7
@@ -31,7 +31,7 @@ EXPOSE 3306
 # This image must forever use UID 27 for mysql user so our volumes are
 # safe in the future. This should *never* change, the last test is there
 # to make sure of that.
-RUN yum install -y yum-utils gettext hostname && \
+RUN yum install -y yum-utils && \
     yum-config-manager --enable rhel-server-rhscl-7-rpms && \
     yum-config-manager --enable rhel-7-server-optional-rpms && \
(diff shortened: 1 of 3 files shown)


mnagy:
@rhcarvalho PTAL

[test]

rhcarvalho:
LGTM

mnagy:
one moment, found other stuff

mnagy:
Made changes to 5.6 Dockerfile to align with that of 5.5. It should not introduce any functional difference, but would be hell to maintain if we want to deduplicate and generate.

PTAL

5.6/Dockerfile.rhel7
```diff
@@ -31,7 +31,7 @@ EXPOSE 3306
 # This image must forever use UID 27 for mysql user so our volumes are
 # safe in the future. This should *never* change, the last test is there
 # to make sure of that.
-RUN yum install -y yum-utils gettext hostname && \
+RUN yum install -y yum-utils && \
```
rhcarvalho: hmmm this was marked for install twice :)

rhcarvalho:
still LGTM

5.6/Dockerfile.rhel7
mnagy: indeed :)

mnagy:
Thanks @rhcarvalho, will merge once test passes.

PR #143, opened 2016-05-19
[DO NOT MERGE] Proof-of-concept deduplication

mnagy:
This is my first shot at an deduplication idea of using sed to generate trees. So far there is only one script in `hack/generate.sh` that, when run from the root directory will generate directories v5.5 and v5.6 from the directory origin, using the appropriate sed files to substitute correct versions and so on.

One thing I don't like is the `if [ "@@REPLICATION@@" = "GTID" ]; then` approach that I'd taken so far. I'd much less be in favour of a macro-style conditional `#ifdef` construct, I'd like to hear discussion on that.

This PR still retains 5.5 and 5.6 directories and is just for initial discussion.

@mfojtik @bparees @hhorak @praiskup @ppisar @pkajaba PTAL

### DIFF
--- a/hack/generate.sh
+++ b/hack/generate.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+IN=origin
+for version in v5.5 v5.6; do
+  cp -r $IN $version
+  find $version -type f -exec sed -i -f "${version}.sed" \{\} \;
+done
--- a/v5.5.sed
+++ b/v5.5.sed
@@ -1,3 +1,6 @@
 s/@@VERSION@@/5.5/
 s|@@IMAGE@@|openshift/mysql-55-centos7|
+s|@@RHEL_IMAGE@@|openshift3/mysql-55-rhel7|
 s/@@REPLICATION@@/manual/
+s/@@COLLECTION@@/mysql55/
+s/@@VERSIONED_TAGS@@/mysql55/
--- a/v5.5.sed
+++ b/v5.5.sed
@@ -1,2 +1,3 @@
 s/@@VERSION@@/5.5/
 s|@@IMAGE@@|openshift/mysql-55-centos7|
+s/@@REPLICATION@@/manual/
--- a/v5.5.sed
+++ b/v5.5.sed
@@ -1,0 +1,2 @@
+s/@@VERSION@@/5.5/
+s|@@IMAGE@@|openshift/mysql-55-centos7|
--- a/v5.5.sed
+++ b/v5.5.sed
@@ -1,6 +1,6 @@
-s/@@VERSION@@/5.5/
-s|@@IMAGE@@|openshift/mysql-55-centos7|
-s|@@RHEL_IMAGE@@|openshift3/mysql-55-rhel7|
-s/@@REPLICATION@@/manual/
-s/@@COLLECTION@@/mysql55/
-s/@@VERSIONED_TAGS@@/mysql55/
+s/@@VERSION@@/5.5/g
+s|@@IMAGE@@|openshift/mysql-55-centos7|g
+s|@@RHEL_IMAGE@@|openshift3/mysql-55-rhel7|g
+s/@@REPLICATION@@/manual/g
+s/@@COLLECTION@@/mysql55/g
+s/@@VERSIONED_TAGS@@/mysql55/g
--- a/origin/Dockerfile
+++ b/origin/Dockerfile
@@ -18,7 +18,7 @@ ENV MYSQL_VERSION=@@VERSION@@ \
 LABEL io.k8s.description="MySQL is a multi-user, multi-threaded SQL database server" \
       io.k8s.display-name="MySQL @@VERSION@@" \
       io.openshift.expose-services="3306:mysql" \
-      io.openshift.tags="database,mysql,mysql56,rh-mysql56"
+      io.openshift.tags="database,mysql,@@VERSIONED_TAGS@@"
 
 EXPOSE 3306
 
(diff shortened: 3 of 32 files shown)


mnagy:
And I forgot @rhcarvalho :)

hack/generate.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+IN=origin
+for version in v5.5 v5.6; do
+  cp -r $IN $version
```
ppisar: Here you should prune $version directory before copying there. It's necessary when you decide to remove a file from the origin directory.

mnagy:
If you want to get the idea of what this does, run hack/generate.sh from the root repository directory and take a look at differences of `mold 5.5 5.6` and `mold v5.5 v5.6` or `mold 5.5 v5.5` and `mold 5.6 v5.6`.

As we talked about on the meeting, ideally we'd have generate.sh be called from a `make update` and write directly into 5.5 and 5.6 directories. Jenkins could then run make update as well and test whether that changes anything in order to make sure we keep consistency.

hack/generate.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+IN=origin
+for version in v5.5 v5.6; do
```
ppisar: Here you could generate the list of version by parsing expanded *.sed glob.

v5.5.sed
```diff
@@ -0,0 +1,6 @@
+s/@@VERSION@@/5.5/
+s|@@IMAGE@@|openshift/mysql-55-centos7|
```
rhcarvalho: you could consistently use the same separator symbol, e.g., ",": `s,@@IMAGE@@,...,`

hack/generate.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+IN=origin
```
rhcarvalho: Calling it `origin` unnecessarily puts my mind to think of openshift/origin...
How about putting the "templates" in rooted at the repo root, or give it a different name? `template/`?

rhcarvalho:
@mnagy thanks for kicking us off with this!

Using the sed approach, do you think we could double check that all lines in the *.sed files end up being used?
I see a possible bug if we make a typo in the @@ZZZ@@ name and notice it too late :)

v5.5.sed
```diff
@@ -0,0 +1,2 @@
+s/@@VERSION@@/5.5/
```
ppisar: The issue with this simple script is what if you need literal @@VERSION@@ somewhere in the generated output. What if the result of a rule is a string that will be be matched by another rule by an accident?

While this seems pretty improbable, it creates a minefield that can hurt you in the future. I still think a real template system is necessary.

ppisar: And you probably want global modifier (s///g).

mnagy:
@rhcarvalho how about instead making sure there are no `@@<anything>@@` left-over?

v5.5.sed
mnagy: I'd pass that bridge when we get there, but I do not think we will ever get there. IMO `@@XYZ@@` is pretty much a standard for substitutions simple as these. If we end up needing something different, it won't be hard to run sed on the whore repository and changing that. But again, I think this will be enough for us. At least I'm hoping python won't be needing something super-complicated with it's 4 versions :)

praiskup:
+1 to this, even though there would be more flexible approaches.

> I'd much less be in favour of a macro-style conditional #ifdef construct,

We don't deal with C-preprocessor, with e.g. jinja people (can) read only generated source (not the preprocessor conditionals).  So basically sed is equivalent to any other substitution method, macro, jinja, or whatever.  IOW - having run-time if-conditionals in your approach is small -1 from me, if end-user opens the script in generated container, it just complicates reading.

I don't know why the double '@@' symbol.  Everywhere (thousands of projects using autotools
use just '@wrapper@', so this looks like non-standard a bit.

More discussions for such trivial tasks are not worth, so I stop here :).

mnagy:
@praiskup I think I will actually change it to get rid of run-time if-conditionals. The `#ifdef` was just an example, I didn't mean to use the exact same syntax. I'm still thinking about how to do it, but I have a general idea.

Hmm, I thought autotools used double @ :) But yeah, I don't think it matters much.

rhcarvalho:
> how about instead making sure there are no @@<anything>@@ left-over?

might be enough to help us fight typos.

praiskup:
> I'm still thinking about how to do it, but I have a general idea.

This sounds like reinventing something.  Have a look how automake.git uses sed (together
with 'make' variable expansion) to generate scripts in parallel (instead of doing find -exec .. sth
which is ugly and not effective):
https://www.gnu.org/software/automake/manual/html_node/Scripts.html

praiskup:
@ppisar wrote:
> While this seems pretty improbable, it creates a minefield that can hurt you in the future. I still think a real template system is necessary.

I agree here 100%.

On the other hand, I bet we still don't try to find the _best_ solution, but rather the quicker.. (that is
why initially the C&P solution was chosen?) and sed is probably good enough.

mnagy:
About the templating system.. Yes, it would be better and safer, but it would require people that want to contribute to install that system. IIRC, we agreed that's the reason we do not want to go that path. For our purposes, I think the sed solution is pretty much enough and the risks we'll need something more is both low and easy to mitigate if we'll really need to.

hhorak:
I would rather avoid complications with `#ifdef` and just stick with simple `sed`s. Anyway, I like the proposed solution, and yes, the check for unexpanded `@`s would be fine.

praiskup:
Works for me, that's why I'm +1, this is about philosophy.

> Yes, it would be better and safer, but it would require people that want to contribute to install that system.

I don't buy this argument.  We try to "protect" contributors who have problems to install
template system (and also those problems can be removed), while we are making the project
boring to hackers.

hack/generate.sh
mnagy: That sounds dangerous. We already have a versions variable in Makefile, I'll use that.

PR #145, opened 2016-05-20
Remove teardown.sh

rhcarvalho:
This script shows it's age and is not needed.

- It refers to RCs when we use DCs;
- It hard codes the service name when 'mysql-master' is just the default
  value of the MYSQL_MASTER_SERVICE_NAME template parameter;
- It doesn't clean up all objects created by the template, as its name
  may suggest.

@mnagy @bparees PTAL


bparees:
lgtm

rhcarvalho:
green tests

PR #146, opened 2016-05-23
Remove useless fields from template

rhcarvalho:
The template is complex enough with the information it needs to have.
Removing fields that were added automatically when bumping version with
an automated tool makes it a bit saner to maintain.

The fields being removed are all specifying default values.


bparees:
lgtm, thanks @rhcarvalho!

PR #147, opened 2016-05-24
Remove debugging volume definition

rhcarvalho:
The PersistentVolume spec defined is only useful in some very particular
environment, with a certain NFS mount at a fixed IP.

The file(s) is(are) not referenced to within this repo.

The `volume.json` file was originally added in #68. @mfojtik any objection to removing it?
@bparees PTAL.


mnagy:
LGTM

bparees:
lgtm

PR #148, opened 2016-05-24
Improve replication template params

rhcarvalho:
Improve descriptions

### DIFF
--- a/5.5/examples/replica/mysql_replica.json
+++ b/5.5/examples/replica/mysql_replica.json
@@ -22,29 +22,29 @@
     },
     {
       "name": "MYSQL_MASTER_PASSWORD",
-      "description": "The password for the MySQL master user",
+      "description": "The password used for the MYSQL_MASTER_USER",
       "generate": "expression",
       "from": "[a-zA-Z0-9]{12}"
     },
     {
       "name": "MYSQL_USER",
-      "description": "The username that clients will use to connect to MySQL server",
+      "description": "The username that clients will use to connect to the server",
       "value": "user"
     },
     {
       "name": "MYSQL_PASSWORD",
-      "description": "The password for the MySQL master user",
+      "description": "The password for the MYSQL_USER",
       "generate": "expression",
       "from": "[a-zA-Z0-9]{12}"
     },
     {
       "name": "MYSQL_DATABASE",
-      "description": "The name of the database that will be created",
+      "description": "The name of the database that will be created and replicated",
       "value": "userdb"
     },
     {
       "name": "MYSQL_ROOT_PASSWORD",
-      "description": "The password for the MySQL adminitrator",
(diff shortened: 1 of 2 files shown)


5.5/examples/replica/mysql_replica.json
```diff
@@ -16,35 +16,35 @@
...
     },
     {
       "name": "MYSQL_USER",
-      "description": "The username that clients will use to connect to MySQL server",
+      "description": "The username that clients will use to connect to the server",
       "value": "user"
     },
     {
       "name": "MYSQL_PASSWORD",
-      "description": "The password for the MySQL master user",
```
rhcarvalho: This was incorrect, `MYSQL_PASSWORD` is not related to the "master user".

5.5/examples/replica/mysql_replica.json
```diff
@@ -16,35 +16,35 @@
...
     },
     {
       "name": "MYSQL_DATABASE",
-      "description": "The name of the database that will be created",
+      "description": "The name of the database that will be created and replicated",
       "value": "userdb"
     },
     {
       "name": "MYSQL_ROOT_PASSWORD",
-      "description": "The password for the MySQL adminitrator",
```
rhcarvalho: There was a typo here. Since the user name is 'root', I think it is clearer to say "root user".
(it took me some time to figure out what was the username for the admin user -- root)

bparees:
lgtm

5.5/examples/replica/mysql_replica.json
```diff
@@ -158,11 +158,11 @@
                 "env": [
                   {
                     "name": "MYSQL_MASTER_USER",
```
rhcarvalho: At first I'm not changing the env var, for it would change the "api" of the image... not sure if it is worth changing the parameter name and have it different than the env var.

bparees: i don't think it's worth changing it, but i'm also not clear why even the parameter name really needed to be changed.

rhcarvalho: It doesn't, really. I thought `MYSQL_REPLICATION_USER` is clearer than `MYSQL_MASTER_USER`, but then stepped back on changing the env var. What I wanted is to hear a second opinion.

bparees: I guess i tend to agree that replication_user is clearer, but if we can't change it in the image itself i'm not sure it's worth changing it in the template and then having them out of sync.

perhaps the best thing to do would be to deprecate master_user, and make the image(scripts) support both values?

rhcarvalho: @bparees after our conversation yesterday on IRC I decided to step back on changing the parameter name.

bparees:
lgtm

PR #151, opened 2016-05-26
Add persistent volume claim to slave

rhcarvalho:
This accommodates for slave restarts without needing to re-sync the entire database, as well as having a hot backup of the data.

Changes to run-mysqld-slave are needed for proper pod start up with a persistent volume.

### DIFF
--- a/5.5/examples/replica/README.md
+++ b/5.5/examples/replica/README.md
@@ -25,9 +25,13 @@ This template will start with one MySQL master server and one slave server.
 
 ## Persistent storage
 
-In order to provide the persistent storage for the MySQL master server, the administrator
-of OpenShift needs to create a PersistentVolume that you can claim. This example requires a PersistentVolume of size 512m be available.
-To learn more about how to create PersistentVolume, refer to [OpenShift documentation](https://docs.openshift.org/latest/install_config/persistent_storage/persistent_storage_nfs.html)
+In order to provide persistent storage for MySQL, this example requires two
+persistent volumes of at least 512 MiB each.
+
+The OpenShift cluster administrator needs to create persistent volumes to be
+claimed when the template is instantiated. Refer to the [OpenShift
+documentation](https://docs.openshift.org/latest/install_config/persistent_storage/persistent_storage_nfs.html)
+to learn how to create persistent volumes.
 
 ### Service 'mysql-master'
 
--- a/5.5/examples/replica/mysql_replica.json
+++ b/5.5/examples/replica/mysql_replica.json
@@ -47,6 +47,13 @@
       "description": "The password for the root user",
       "generate": "expression",
       "from": "[a-zA-Z0-9]{12}"
+    },
+    {
+      "name": "VOLUME_CAPACITY",
+      "displayName": "Volume Capacity",
+      "description": "Volume space available for data, e.g. 512Mi, 2Gi.",
+      "value": "1Gi",
+      "required": true
     }
   ],
   "objects": [
@@ -62,7 +69,24 @@
         ],
         "resources": {
           "requests": {
-            "storage": "512Mi"
+            "storage": "${VOLUME_CAPACITY}"
+          }
+        }
+      }
+    },
+    {
+      "kind": "PersistentVolumeClaim",
+      "apiVersion": "v1",
+      "metadata": {
+        "name": "mysql-slave"
+      },
+      "spec": {
+        "accessModes": [
+          "ReadWriteOnce"
+        ],
+        "resources": {
+          "requests": {
+            "storage": "${VOLUME_CAPACITY}"
           }
         }
       }
@@ -220,6 +244,14 @@
             }
           },
           "spec": {
+            "volumes": [
+              {
+                "name": "mysql-slave-data",
+                "persistentVolumeClaim": {
+                  "claimName": "mysql-slave"
+                }
+              }
+            ],
             "containers": [
               {
                 "name": "server",
@@ -248,6 +280,12 @@
                   {
                     "name": "MYSQL_DATABASE",
                     "value": "${MYSQL_DATABASE}"
+                  }
+                ],
+                "volumeMounts": [
+                  {
+                    "name": "mysql-slave-data",
+                    "mountPath": "/var/lib/mysql/data"
                   }
                 ],
                 "imagePullPolicy": "IfNotPresent"
--- a/5.5/root/usr/bin/run-mysqld-slave
+++ b/5.5/root/usr/bin/run-mysqld-slave
@@ -10,12 +10,6 @@ export_setting_variables
 
 log_volume_info $MYSQL_DATADIR
 
-# Just run normal server if the data directory is already initialized
-if [ -d "${MYSQL_DATADIR}/mysql" ]; then
-  log_info "Datadir already exists, execing ordinary run script"
-  exec /usr/bin/run-mysqld "$@"
-fi
-
 export MYSQL_RUNNING_AS_SLAVE=1
 
 [ -f ${CONTAINER_SCRIPTS_PATH}/validate_replication_variables.sh ] && source ${CONTAINER_SCRIPTS_PATH}/validate_replication_variables.sh
@@ -31,31 +25,34 @@ envsubst < ${CONTAINER_SCRIPTS_PATH}/my-paas.cnf.template > /etc/my.cnf.d/paas.c
 envsubst < ${CONTAINER_SCRIPTS_PATH}/my-slave.cnf.template > /etc/my.cnf.d/slave.cnf
 envsubst < ${CONTAINER_SCRIPTS_PATH}/my-tuning.cnf.template > /etc/my.cnf.d/tuning.cnf
 
-# Initialize MySQL database and wait for the MySQL master to accept
-# connections.
-initialize_database "$@"
-wait_for_mysql_master
+if [ ! -e "${MYSQL_DATADIR}/mysql" ]; then
+  # Initialize MySQL database and wait for the MySQL master to accept
+  # connections.
+  initialize_database "$@"
+  wait_for_mysql_master
 
-# Get binlog file and position from master
-STATUS_INFO=$(mysql --host "$MYSQL_MASTER_SERVICE_NAME" "-u${MYSQL_MASTER_USER}" "-p${MYSQL_MASTER_PASSWORD}" replication -e 'SELECT File, Position from replication\G')
-BINLOG_POSITION=$(echo "$STATUS_INFO" | grep 'Position:' | head -n 1 | sed -e 's/^\s*Position: //')
-BINLOG_FILE=$(echo "$STATUS_INFO" | grep 'File:' | head -n 1 | sed -e 's/^\s*File: //')
+  # Get binlog file and position from master
+  STATUS_INFO=$(mysql --host "$MYSQL_MASTER_SERVICE_NAME" "-u${MYSQL_MASTER_USER}" "-p${MYSQL_MASTER_PASSWORD}" replication -e 'SELECT File, Position from replication\G')
+  BINLOG_POSITION=$(echo "$STATUS_INFO" | grep 'Position:' | head -n 1 | sed -e 's/^\s*Position: //')
+  BINLOG_FILE=$(echo "$STATUS_INFO" | grep 'File:' | head -n 1 | sed -e 's/^\s*File: //')
 
-if [ -z "${BINLOG_FILE}" -o -z "${BINLOG_POSITION}" ] ; then
-  echo "Could not read binlog position or file from master"
-  exit 1
-fi
+  if [ -z "${BINLOG_FILE}" -o -z "${BINLOG_POSITION}" ] ; then
+    echo "Could not read binlog position or file from master"
+    exit 1
+  fi
 
-mysql $mysql_flags <<EOSQL
-  CHANGE MASTER TO MASTER_HOST='${MYSQL_MASTER_SERVICE_NAME}',MASTER_USER='${MYSQL_MASTER_USER}', MASTER_PASSWORD='${MYSQL_MASTER_PASSWORD}', MASTER_LOG_FILE='${BINLOG_FILE}', MASTER_LOG_POS=${BINLOG_POSITION};
-  SET GLOBAL SQL_SLAVE_SKIP_COUNTER = 1; SLAVE START;
+  mysql $mysql_flags <<EOSQL
+CHANGE MASTER TO MASTER_HOST='${MYSQL_MASTER_SERVICE_NAME}',MASTER_USER='${MYSQL_MASTER_USER}', MASTER_PASSWORD='${MYSQL_MASTER_PASSWORD}', MASTER_LOG_FILE='${BINLOG_FILE}', MASTER_LOG_POS=${BINLOG_POSITION};
+SET GLOBAL SQL_SLAVE_SKIP_COUNTER = 1; SLAVE START;
 EOSQL
 
-log_info 'Sourcing post-init.sh ...'
-[ -f ${CONTAINER_SCRIPTS_PATH}/post-init.sh ] && source ${CONTAINER_SCRIPTS_PATH}/post-init.sh
+  log_info 'Sourcing post-init.sh ...'
+  [ -f ${CONTAINER_SCRIPTS_PATH}/post-init.sh ] && source ${CONTAINER_SCRIPTS_PATH}/post-init.sh
 
-# Restart the MySQL server with public IP bindings
-shutdown_local_mysql
+  # Restart the MySQL server with public IP bindings
+  shutdown_local_mysql
+fi
+
 unset_env_vars
 log_volume_info $MYSQL_DATADIR
 log_info 'Running final exec -- Only MySQL server logs after this point'
--- a/5.6/examples/replica/mysql_replica.json
+++ b/5.6/examples/replica/mysql_replica.json
@@ -47,6 +47,13 @@
       "description": "The password for the root user",
       "generate": "expression",
       "from": "[a-zA-Z0-9]{12}"
+    },
+    {
+      "name": "VOLUME_CAPACITY",
+      "displayName": "Volume Capacity",
+      "description": "Volume space available for data, e.g. 512Mi, 2Gi.",
+      "value": "1Gi",
+      "required": true
     }
   ],
   "objects": [
@@ -62,7 +69,24 @@
         ],
         "resources": {
           "requests": {
-            "storage": "512Mi"
+            "storage": "${VOLUME_CAPACITY}"
+          }
+        }
+      }
+    },
+    {
+      "kind": "PersistentVolumeClaim",
+      "apiVersion": "v1",
+      "metadata": {
+        "name": "mysql-slave"
+      },
+      "spec": {
+        "accessModes": [
+          "ReadWriteOnce"
+        ],
+        "resources": {
+          "requests": {
+            "storage": "${VOLUME_CAPACITY}"
           }
         }
       }
@@ -220,6 +244,14 @@
             }
           },
           "spec": {
+            "volumes": [
+              {
+                "name": "mysql-slave-data",
+                "persistentVolumeClaim": {
+                  "claimName": "mysql-slave"
+                }
+              }
+            ],
             "containers": [
               {
                 "name": "server",
@@ -248,6 +280,12 @@
                   {
                     "name": "MYSQL_DATABASE",
                     "value": "${MYSQL_DATABASE}"
+                  }
+                ],
+                "volumeMounts": [
+                  {
+                    "name": "mysql-slave-data",
+                    "mountPath": "/var/lib/mysql/data"
                   }
                 ],
                 "imagePullPolicy": "IfNotPresent"
--- a/5.6/examples/replica/README.md
+++ b/5.6/examples/replica/README.md
@@ -25,9 +25,13 @@ This template will start with one MySQL master server and one slave server.
 
 ## Persistent storage
 
-In order to provide the persistent storage for the MySQL master server, the administrator
-of OpenShift needs to create a PersistentVolume that you can claim. This example requires a PersistentVolume of size 512m be available.
-To learn more about how to create PersistentVolume, refer to [OpenShift documentation](https://docs.openshift.org/latest/install_config/persistent_storage/persistent_storage_nfs.html)
+In order to provide persistent storage for MySQL, this example requires two
+persistent volumes of at least 512 MiB each.
+
+The OpenShift cluster administrator needs to create persistent volumes to be
+claimed when the template is instantiated. Refer to the [OpenShift
+documentation](https://docs.openshift.org/latest/install_config/persistent_storage/persistent_storage_nfs.html)
+to learn how to create persistent volumes.
 
 ### Service 'mysql-master'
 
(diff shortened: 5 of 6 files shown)


rhcarvalho:
@mnagy @bparees @hhorak PTAL

hhorak:
FYI @misli

5.5/root/usr/bin/run-mysqld-slave
```diff
@@ -10,12 +10,6 @@ export_setting_variables
 
 log_volume_info $MYSQL_DATADIR
 
-# Just run normal server if the data directory is already initialized
-if [ -d "${MYSQL_DATADIR}/mysql" ]; then
-  log_info "Datadir already exists, execing ordinary run script"
-  exec /usr/bin/run-mysqld "$@"
```
rhcarvalho: We cannot do this because this is still a "slave" pod. It doesn't have the env vars required to run using `run-mysqld`.

With this in, after restarting the slave pod we have a crash loop with this log and exit code 1:

```
---> 10:07:51     Datadir already exists, execing ordinary run script
You must either specify the following environment variables:
  MYSQL_USER (regex: '^[a-zA-Z0-9_]+$')
  MYSQL_PASSWORD (regex: '^[a-zA-Z0-9_~!@#$%^&*()-=<>,.?;:|]+$')
  MYSQL_DATABASE (regex: '^[a-zA-Z0-9_]+$')
Or the following environment variable:
  MYSQL_ROOT_PASSWORD (regex: '^[a-zA-Z0-9_~!@#$%^&*()-=<>,.?;:|]+$')
Or both.
Optional Settings:
  MYSQL_LOWER_CASE_TABLE_NAMES (default: 0)
  MYSQL_MAX_CONNECTIONS (default: 151)
  MYSQL_FT_MIN_WORD_LEN (default: 4)
  MYSQL_FT_MAX_WORD_LEN (default: 20)
  MYSQL_AIO (default: 1)
  MYSQL_KEY_BUFFER_SIZE (default: 32M or 10% of available memory)
  MYSQL_MAX_ALLOWED_PACKET (default: 200M)
  MYSQL_TABLE_OPEN_CACHE (default: 400)
  MYSQL_SORT_BUFFER_SIZE (default: 256K)
  MYSQL_READ_BUFFER_SIZE (default: 8M or 5% of available memory)
  MYSQL_INNODB_BUFFER_POOL_SIZE (default: 32M or 50% of available memory)
  MYSQL_INNODB_LOG_FILE_SIZE (default: 8M or 15% of available memory)
  MYSQL_INNODB_LOG_BUFFER_SIZE (default: 8M or 15% of available memory)
```

`MYSQL_USER` and `MYSQL_PASSWORD` are not available to the slave pod.


Instead, as @mnagy proposed, we're wrapping the initialization steps in a `if` that checks that the directory doesn't exist.

hhorak:
The changes make sense to me, although I haven't checked the template changes, since I don't understand that piece much. Anyway, LGTM.

rhcarvalho:
@hhorak thanks for reviewing.

mnagy:
LGTM

Have you ran any extended tests with openshift?

rhcarvalho:
@mnagy no, but the tested is commented out anyway, isn't it?!

Is it possible to run [test] here?

5.5/root/usr/bin/run-mysqld-slave
```diff
@@ -31,31 +25,34 @@ envsubst < ${CONTAINER_SCRIPTS_PATH}/my-paas.cnf.template > /etc/my.cnf.d/paas.c
 envsubst < ${CONTAINER_SCRIPTS_PATH}/my-slave.cnf.template > /etc/my.cnf.d/slave.cnf
 envsubst < ${CONTAINER_SCRIPTS_PATH}/my-tuning.cnf.template > /etc/my.cnf.d/tuning.cnf
 
-# Initialize MySQL database and wait for the MySQL master to accept
-# connections.
-initialize_database "$@"
-wait_for_mysql_master
+if [ ! -e "${MYSQL_DATADIR}/mysql" ]; then
```
bparees: so if this file does exist, (because it's a restart) the pod will start up and join as a normal slave?

bparees:
> This accomodates for slave restarts without needing to re-sync the
entire database.

well, this is a debatable change.  rsyncing is certainly undesirable, but so is allocating a PV to each slave...  our original intent was that the master PV is the system of record, and the slaves should be epheremeral/lightweight and just replicate on startup.  It's especially useful to limit PV usage for online since that's limited to 2 PVs per project.

It seems like this is something users might want to choose (allocate a PV and get faster slave restart, don't allocate a PV and get lighter weight slaves)

5.5/root/usr/bin/run-mysqld-slave
rhcarvalho: > so if this file does exist, (because it's a restart) the pod will start up and join as a normal slave?

If it's a restart, then the directory DOES exist.

In any case, we should always start a container with entrypoint "run-mysql-slave" as a slave, see https://github.com/openshift/mysql/pull/151#discussion_r64735585.

bparees: > If it's a restart, then the directory DOES exist.

isn't that what i said? 

i'm trying to confirm that when a slave restarts and the PV already contains old data, the slave will just pick up where it left off, rejoining as a slave w/ no need to go through any of the initialize_db/wait_for_master/etc steps that it went through when the PV was empty.

rhcarvalho: Sorry I misunderstood your sentence at first.

I didn't do extensive testing with loads of data, but just went far enough to verify that it works.


The slave log after restart:

```
--> 11:11:12     The 'slave' server-id is 1657242886
---> 11:11:12     Processing MySQL configuration files ...
---> 11:11:12     Cleaning up environment variables MYSQL_USER, MYSQL_PASSWORD, MYSQL_DATABASE and MYSQL_ROOT_PASSWORD ...
---> 11:11:12     Running final exec -- Only MySQL server logs after this point
160526 11:11:12 [Note] /opt/rh/mysql55/root/usr/libexec/mysqld (mysqld 5.5.45-log) starting as process 1 ...
160526 11:11:12 [Note] Plugin 'FEDERATED' is disabled.
160526 11:11:12 InnoDB: The InnoDB memory heap is disabled
160526 11:11:12 InnoDB: Mutexes and rw_locks use GCC atomic builtins
160526 11:11:12 InnoDB: Compressed tables use zlib 1.2.7
160526 11:11:12 InnoDB: Using Linux native AIO
160526 11:11:12 InnoDB: Initializing buffer pool, size = 32.0M
160526 11:11:12 InnoDB: Completed initialization of buffer pool
160526 11:11:12 InnoDB: highest supported file format is Barracuda.
InnoDB: Log scan progressed past the checkpoint lsn 1701194
160526 11:11:12  InnoDB: Database was not shut down normally!
InnoDB: Starting crash recovery.
InnoDB: Reading tablespace information from the .ibd files...
InnoDB: Restoring possible half-written data pages from the doublewrite
InnoDB: buffer...
InnoDB: Doing recovery: scanned up to log sequence number 1701204
InnoDB: Last MySQL binlog file position 0 107, file name /var/lib/mysql/data/mysql-bin.000004
160526 11:11:12  InnoDB: Waiting for the background threads to start
160526 11:11:13 InnoDB: 5.5.45 started; log sequence number 1701204
160526 11:11:13 [Note] Recovering after a crash using /var/lib/mysql/data/mysql-bin
160526 11:11:13 [Note] Starting crash recovery...
160526 11:11:13 [Note] Crash recovery finished.
160526 11:11:13 [Note] Server hostname (bind-address): '0.0.0.0'; port: 3306
160526 11:11:13 [Note]   - '0.0.0.0' resolves to '0.0.0.0';
160526 11:11:13 [Note] Server socket created on IP: '0.0.0.0'.
160526 11:11:13 [Warning] 'user' entry 'root@mysql-slave-1-83ipc' ignored in --skip-name-resolve mode.
160526 11:11:13 [Warning] 'user' entry '@mysql-slave-1-83ipc' ignored in --skip-name-resolve mode.
160526 11:11:13 [Warning] 'proxies_priv' entry '@ root@mysql-slave-1-83ipc' ignored in --skip-name-resolve mode.
160526 11:11:13 [Note] Event Scheduler: Loaded 0 events
160526 11:11:13 [Note] /opt/rh/mysql55/root/usr/libexec/mysqld: ready for connections.
Version: '5.5.45-log'  socket: '/var/lib/mysql/mysql.sock'  port: 3306  MySQL Community Server (GPL)
160526 11:11:13 [Note] Slave SQL thread initialized, starting replication in log 'mysql-bin.000004' at position 27399, relay log '/var/lib/mysql/data/mysql-relay-bin.000057' position: 253
160526 11:11:13 [Note] Slave I/O thread: connected to master 'master@mysql-master:3306',replication started in log 'mysql-bin.000004' at position 27399
```

... suggests that it picked up from where it left.

Then, I wrote more data in the master and it was successfully synced to the slave.


In a separate restart cycle, as I write this comment, I decided to write data to master while the slave is down and ensure that it gets synced when the slave comes back up.
Result: items written to master while slave was down were successfully synced after restart.

Log says:

```
---> 15:23:30     The 'slave' server-id is 1657242886
---> 15:23:30     Processing MySQL configuration files ...
---> 15:23:30     Cleaning up environment variables MYSQL_USER, MYSQL_PASSWORD, MYSQL_DATABASE and MYSQL_ROOT_PASSWORD ...
---> 15:23:30     Running final exec -- Only MySQL server logs after this point
160526 15:23:30 [Note] /opt/rh/mysql55/root/usr/libexec/mysqld (mysqld 5.5.45-log) starting as process 1 ...
160526 15:23:30 [Note] Plugin 'FEDERATED' is disabled.
160526 15:23:30 InnoDB: The InnoDB memory heap is disabled
160526 15:23:30 InnoDB: Mutexes and rw_locks use GCC atomic builtins
160526 15:23:30 InnoDB: Compressed tables use zlib 1.2.7
160526 15:23:30 InnoDB: Using Linux native AIO
160526 15:23:30 InnoDB: Initializing buffer pool, size = 32.0M
160526 15:23:30 InnoDB: Completed initialization of buffer pool
160526 15:23:30 InnoDB: highest supported file format is Barracuda.
160526 15:23:30  InnoDB: Waiting for the background threads to start
160526 15:23:31 InnoDB: 5.5.45 started; log sequence number 1701412
160526 15:23:31 [Note] Server hostname (bind-address): '0.0.0.0'; port: 3306
160526 15:23:31 [Note]   - '0.0.0.0' resolves to '0.0.0.0';
160526 15:23:31 [Note] Server socket created on IP: '0.0.0.0'.
160526 15:23:31 [Warning] 'user' entry 'root@mysql-slave-1-83ipc' ignored in --skip-name-resolve mode.
160526 15:23:31 [Warning] 'user' entry '@mysql-slave-1-83ipc' ignored in --skip-name-resolve mode.
160526 15:23:31 [Warning] 'proxies_priv' entry '@ root@mysql-slave-1-83ipc' ignored in --skip-name-resolve mode.
160526 15:23:31 [Note] Slave SQL thread initialized, starting replication in log 'mysql-bin.000004' at position 27629, relay log '/var/lib/mysql/data/mysql-relay-bin.000059' position: 483
160526 15:23:31 [Note] Event Scheduler: Loaded 0 events
160526 15:23:31 [Note] /opt/rh/mysql55/root/usr/libexec/mysqld: ready for connections.
Version: '5.5.45-log'  socket: '/var/lib/mysql/mysql.sock'  port: 3306  MySQL Community Server (GPL)
160526 15:23:31 [Note] Slave I/O thread: connected to master 'master@mysql-master:3306',replication started in log 'mysql-bin.000004' at position 27629
```


Notice that in the last line the replication is started from different positions at the binlog.

bparees: cool, thanks for the thorough vetting!

bparees:
@rhcarvalho i think i'm ok w/ switching this to have slaves use a PV, but can you update the readme to let people know that
1) the slaves use a PV to optimize restart/resync performance
2) if they want to use an emptydir volume for the slaves(not the master), it will still work correctly (but startup performance will suffer).

?

5.5/examples/replica/README.md
```diff
@@ -26,8 +26,12 @@ This template will start with one MySQL master server and one slave server.
...
 
 In order to provide the persistent storage for the MySQL master server, the administrator
-of OpenShift needs to create a PersistentVolume that you can claim. This example requires a PersistentVolume of size 512m be available.
-To learn more about how to create PersistentVolume, refer to [OpenShift documentation](https://docs.openshift.org/latest/install_config/persistent_storage/persistent_storage_nfs.html)
+of OpenShift needs to create a PersistentVolume that you can claim. This example requires two persistent volumes of size 512m be available.
+To learn more about how to create PersistentVolume, refer to the [OpenShift documentation](https://docs.openshift.org/latest/install_config/persistent_storage/persistent_storage_nfs.html).
+
+The slave use persistent storage to serve as a hot backup and to optimize
+restart performance, such that only the data written while the slave was down
+needs to be replicated and not the entire database.
```
rhcarvalho: @bparees added this note as you asked.

rhcarvalho:
@bparees added a note for point (1), but not for (2) because that one is really a general possibility for the admin. If whoever decides to use an emptyDir volume, he must be coming from the OpenShift docs with the knowledge. In theory, one can use emptyDir for any volume he wants to, and the consequences are directly related to how emptyDir works.

In fact, I started to write it down but it started to sound like an advice to "use" emptyDir, which I can't really advise for :)

5.5/examples/replica/README.md
```diff
@@ -26,8 +26,12 @@ This template will start with one MySQL master server and one slave server.
 ## Persistent storage
 
 In order to provide the persistent storage for the MySQL master server, the administrator
-of OpenShift needs to create a PersistentVolume that you can claim. This example requires a PersistentVolume of size 512m be available.
-To learn more about how to create PersistentVolume, refer to [OpenShift documentation](https://docs.openshift.org/latest/install_config/persistent_storage/persistent_storage_nfs.html)
+of OpenShift needs to create a PersistentVolume that you can claim. This example requires two persistent volumes of size 512m be available.
+To learn more about how to create PersistentVolume, refer to the [OpenShift documentation](https://docs.openshift.org/latest/install_config/persistent_storage/persistent_storage_nfs.html).
+
+The slave use persistent storage to serve as a hot backup and to optimize
```
bparees: uses

5.5/examples/replica/README.md
bparees: "The slave template can be modified to use an EmptyDir volume to avoid the need for the additional persistent volume, at the cost of lengthier restart times due to the need to replicate the entire database to the slave when it restarts."

bparees:
>  If whoever decides to use an emptyDir volume, he must be coming from the OpenShift docs with the knowledge. In theory, one can use emptyDir for any volume he wants to, and the consequences are directly related to how emptyDir works.

I don't think that's entirely true.  In some cases it's a really bad idea to use emptydir because you'll lose data.  In this case you will not lose data, which is fundamental to how this particular example works (that the slaves will always re-replicate the data on startup).  So i think it's still a clarification worth making, especially since users are limited to 2 PVs in openshift online, so they'll be particularly interested in ways to limit their PV usage.

5.5/examples/replica/README.md
rhcarvalho: - One doesn't need to change the template do use emptyDir, because the template defines a PVC, not a PV.
- In real use scenarios, if you already have data on the master, replication won't work without manual intervention. See the documentation on how to enable replication when the master has initial data: https://dev.mysql.com/doc/refman/5.5/en/replication-howto-existingdata.html. It requires a dump and import. BTW this is not very different than MongoDB.

rhcarvalho:
> the slaves will always re-replicate the data on startup

This is my turn to say this is not entirely true :)


The README for the replication example starts with:

```
**This is only a Proof-Of-Concept example and it is not meant to be used in any
production. Use at your own risk.**
```

This is not what people should be using in OSO. And if you consider resource limitations, people should use the standalone persistent template which is what we offer.

Using master-slave and trying to save on storage resources is really an anti-pattern.

5.5/examples/replica/README.md
rhcarvalho: This paragraph is not very accurate, as the persistent storage is not for "optimizing restart performance" but really as I said in the PR "to accommodate for slave restarts".
If there is data in the master (more than BINLOG size) and no data in the slave, replication won't work.

rhcarvalho: And it is weird to have to justify why we use a PVC for the slave, while no mention as to why there is a PVC for master. IMHO it is not the case that one is more obvious than the other.

That README file is really outdated, that's why I opened https://github.com/openshift/mysql/issues/150 to track that.

rhcarvalho:
Reading more of the docs, I am not clear on the behavior when there's data in the master (let's say ~200 MiB out of the 512 claimed?).

Looks like MySQL keeps binlog files around indefinitely until we execute `mysqladmin flush-logs` to remove old logs... So, technically, it would mean that a slave could sync from empty to latest using solely the binlog if all logs are still available. I suspect that's not the most effective way of operating.

Having the logs indefinitely means that master will run out of disk space much "earlier" because of the old logs taking up space...

Our replication example needs a sysadmin :)

https://dev.mysql.com/doc/refman/5.5/en/binary-log.html

And once a manual intervention would clean up old logs, restarting a slave would be impossible without persistent storage or manual intervention.

@bparees since we don't officially support this replication template, and AFAIK don't battle test it to understand how it behaves under normal usage conditions, I think it is not a great usage of time to try to document little details we're not even sure about.

5.5/examples/replica/README.md
bparees: > One doesn't need to change the template do use emptyDir, because the template defines a PVC, not a PV.

I don't understand what your statement here...  PVC and EmptyDir are mutually exclusive.  To use an emptydir instead of a PV, you need to remove the PVC and add an EmptyDir:

https://github.com/openshift/origin/blob/master/examples/db-templates/mongodb-persistent-template.json#L164-L170

vs

https://github.com/openshift/origin/blob/master/examples/db-templates/mongodb-ephemeral-template.json#L147-L153

so you either need to edit the template, or edit the DC after creating it.  Editing the template would be preferred.  You don't define an "emptydir PV" and then use a PVC to claim it.

> In real use scenarios, if you already have data on the master, replication won't work without manual intervention. See the documentation on how to enable replication when the master has initial data: https://dev.mysql.com/doc/refman/5.5/en/replication-howto-existingdata.html. It requires a dump and import. BTW this is not very different than MongoDB.

I thought we discussed this change and you indicated the only benefit from moving to a PV instead of an emptydir was removing the need to re-do a full replication on startup.  If that's not accurate, then I agree that there may not be value in describing how to do this with an emptydir.  In any case if my interpretation of your link is correct, then it means the existing templates never worked if a slave restarted (unless the db is smaller than the binlog, perhaps).  That could be fixed by updating the slave scripts to perform a remote dump of the master, on startup if the db directory is empty.

> And it is weird to have to justify why we use a PVC for the slave, while no mention as to why there is a PVC for master. IMHO it is not the case that one is more obvious than the other.

It seems obvious to me.  The master needs a PV because it is the store of record, and the slave does not need a PV because it is just a replica which bootstraps from the master when it comes up.  If that assumption is incorrect (that the slave can bootstrap from the master w/o additional intervention) then the original implementation was broken from the start, but could be (should be) fixed by performing the db dump step. It doesn't have to be fixed in this PR, but fixing it would make the mysql replica topology more useful since you could run N slaves for better read performance without needing N persistent volumes.  The other advantage to running slaves on an emptydir instead of a PV is that you can run replicas with a single replication controller, which you can't (currently) do with a PV.

bparees:
> Using master-slave and trying to save on storage resources is really an anti-pattern.

I disagree.  One of the primary values of this setup is that I can have a single store of record (the master) but many lightweight replicas that i can read from, to increase my read parallelism without increasing my (persistent) storage costs.  I don't see what's anti-pattern about that.

> So, technically, it would mean that a slave could sync from empty to latest using solely the binlog if all logs are still available. I suspect that's not the most effective way of operating.

Again, depends on whether restarts/pod movement is frequent in your environment and what else you care about.  It's a trade-off, certainly.  And again, we could update our slave logic to perform a remote db dump to bootstrap instead of needing the full binlog available.

> Having the logs indefinitely means that master will run out of disk space much "earlier" because of the old logs taking up space...
> Our replication example needs a sysadmin :)

agree :)  Hence the scary warnings about it not being production ready.

> And once a manual intervention would clean up old logs, restarting a slave would be impossible without persistent storage or manual intervention.

unless/until we update the slaves to grab a db dump.

> @bparees since we don't officially support this replication template, and AFAIK don't battle test it to understand how it behaves under normal usage conditions, I think it is not a great usage of time to try to document little details we're not even sure about.

Agree, you've convinced me that at a minimum the current implementation is mostly broken if you use an emptydir volume for the slave pod.  I still think that can and should be fixed (master should rotate the binlog automatically and slaves should bootstrap via a db dump), but not in this PR.

5.5/examples/replica/mysql_replica.json
```diff
@@ -68,6 +68,23 @@
...
+      "metadata": {
+        "name": "mysql-slave"
+      },
+      "spec": {
+        "accessModes": [
+          "ReadWriteOnce"
+        ],
+        "resources": {
+          "requests": {
+            "storage": "512Mi"
```
bparees: This needs to be a parameter with a default of 1gig because online doesn't have 512meg volumes, see:
https://github.com/openshift/origin/blob/master/examples/db-templates/mongodb-persistent-template.json#L232-L237
https://github.com/openshift/origin/blob/master/examples/db-templates/mongodb-persistent-template.json#L43-L57

5.6/examples/replica/mysql_replica.json
```diff
@@ -68,6 +68,23 @@
...
+      "metadata": {
+        "name": "mysql-slave"
+      },
+      "spec": {
+        "accessModes": [
+          "ReadWriteOnce"
+        ],
+        "resources": {
+          "requests": {
+            "storage": "512Mi"
```
bparees: same here (make it a parameter, 1 gig default)

bparees:
one nit and lgtm.

5.5/examples/replica/mysql_replica.json
rhcarvalho: Is this template meant to be custom-made for OSO?

The master has a claim of 512 MiB, that's why I picked the same number.
The claim being of 512 MiB doesn't require the PVs to have 512 MiB. The claim is <= PV size.

IOW, a claim of 512 MiB can work with any PV of 512 MiB or more, 1 GiB included.

rhcarvalho:
Jenkins is failing with "Something broken on AWS":

```
There was an error talking to AWS. The error message is shown
below:

Blocked => This account is currently blocked and not recognized as a valid account. Please contact aws-verification@amazon.com if you have questions.
```

5.5/examples/replica/mysql_replica.json
bparees: it's meant to be kept in alignment with the OSO templates where-ever possible, there's no good reason *not* to do so.  (So that means changing the master also.  They should probably both use the same parameter since it would make no sense to have different sized volumes for replicas of the same database).

rhcarvalho:
@bparees made the volume 1Gi as you asked.

bparees:
thanks, squash please and lgtm.

rhcarvalho:
@bparees squashed. Sorry for the delay, I lost track of it.

bparees:
lgtm, will merge after tests complete.  thanks @rhcarvalho

Issue #13, opened 2015-03-04
More verbose logging


mnagy:
We would probably like to log queries for example.

mfojtik:
only as special env var, not by default

bparees:
yeah this would be something that should be configurable if it is not already.

soltysh:
@mfojtik +1

hhorak:
logging queries may be quite dangerous, e.g. if anybody stores sensitive data in plain text.. but make it configurable could make sense.

hhorak:
This does not seem to be worth doing it specially, generic way of extending image should be enough. That said, support for extending configuration file passing via ConfigMap will be implemented soon.

PR #153, opened 2016-06-27
Query logging enabled with ENV variable

trepik:
First try to solve #13 
(not tested)


5.5/README.md
```diff
@@ -40,6 +40,8 @@ The following environment variables influence the MySQL configuration file. They
 |  `MYSQL_INNODB_LOG_BUFFER_SIZE` | The size of the buffer that InnoDB uses to write to the log files on disk | 8M (or 15% of available memory)
 |  `MYSQL_DEFAULTS_FILE`          | Point to an alternative configuration file                        |  /etc/my.cnf
 |  `MYSQL_BINLOG_FORMAT`          | Set sets the binlog format, supported values are `row` and `statement` | statement
+|  `MYSQL_LOG_QUERIES`            | Turns on logging of queries | 0
```
bparees: we've generally tried to use true/false (and name it MYSQL_LOG_QUERIES_ENABLED), is "0" and "1" the values mysql requires for its actual configuration to enable this?

bparees:
[test]

5.5/README.md
trepik: "By default, the general query log is disabled. To specify the initial general query log state explicitly, use --general_log[={0|1}]. With no argument or an argument of 1, --general_log enables the log. With an argument of 0, this option disables the log. To specify a log file name, use --general_log_file=file_name."
from [DOC](http://dev.mysql.com/doc/refman/5.7/en/query-log.html)

bparees: ok thanks,  can you add "set to 1 to enable" or something to the description?

bparees:
[test]

bparees:
lgtm.  @hhorak @praiskup any comments?

bparees:
oh and @trepik please squash your commits.

bparees:
(@hhorak you might want to turn on PR merge squashing for all the container repos: https://help.github.com/articles/configuring-pull-request-merge-squashing/)

hhorak:
Looks good to me, LGTM. The squashing is already enabled, but I'm not sure whether it works the way I think it does :) Let's try it..

hhorak:
It looks the squashing works fine, thanks for pointing to that, @bparees.

bparees:
> Looks good to me, LGTM. The squashing is already enabled

weird, I only had a "merge" button and not a "squash+merge" which i was expecting to see.

hhorak:
> weird, I only had a "merge" button and not a "squash+merge" which i was expecting to see.

I was looking for it few seconds as well, there is a possibility to change an action on the page with commit message.

bparees:
ah, thanks!

PR #156, opened 2016-07-05
replace docker-scripts with docker-squash

ryanj:


bparees:
lgtm

omron93:
@bparees Why PR are tested with SKIP_SQUASH=1???

Squashing is not working with this change... (except rh-python35 it is not working for all SCLs and system python in Centos/RHEL 7)

bparees:
answered here (including reference for others who come along): https://github.com/sclorg/mariadb-container/pull/9#issuecomment-238553265

PR #157, opened 2016-07-12
Added cccp.yml files for CentOS Container pipeline.

mohammedzee1000:
This is for the purpose of building these containers with
centos community container pipeline.


hhorak:
This does not influence an image at all, LGTM.

PR #162, opened 2016-10-04
Don't create containers with name during testing

omron93:
mariadb-container and mysql-container contains almost same `test/run` script. So now it is not possible to run mariadb and mysql tests simultaneously (both creates containers `config_test` and `dynamic_config_test`).

This PR fixes this. So test script doesn't create containers with name and it address them by cidfile.

@hhorak @bparees @mnagy Please take a look.


bparees:
lgtm
[test]

PR #165, opened 2016-11-01
remove force tag flag

bparees:


bparees:
[test]

bparees:
[test]

5.6/test/run
```diff
@@ -183,7 +183,7 @@ function assert_login_access() {
 
 function assert_local_access() {
   local id="$1" ; shift
-  docker exec $(get_cid "$id") bash -c mysql <<< "SELECT 1;"
+  docker exec $(get_cid "$id") bash -c 'mysql <<< "SELECT 1;"'
```
praiskup: @bparees What exactly changed in docker?  Is that docker bug (broken STDIN support), or shouldn't we use `docker exec -t`?

bparees: I don't know, but the existing command just hung indefinitely.  I didn't have time to investigate it since i needed it passing so i could get the images rebuilt+pushed.

PR #170, opened 2016-11-15
Show testsuite result explicitly

hhorak:
It was not clear from the first sight from log whether test-suite passed or failed. Adding an explicit message as last line should help.


hhorak:
Changes only in tests and tests passed, so hopefully ok to merge..

PR #171, opened 2016-11-15
Make sources same across 5.6 and 5.7 to allow easier patching

hhorak:
Version 5.5 is already diverged, so I don't see a big point to make it the same there.


hhorak:
[test]

omron93:
[test]

PR #174, opened 2016-12-09
Ignore lost+found directory that exists in mounted persistent volumes

luciddreamz:
When using the MySQL 5.7 image with a PV in OpenShift, MySQL goes into
a crash loop with ‘[ERROR] --initialize specified but the data
directory exists. Aborting.’ Upon investigation, new PV’s have a
lost+found directory, which is permitted in MySQL 5.7.11+, so long as
—ignore-db-dir is set. The issue is described in the MySQL docs here:
https://dev.mysql.com/doc/refman/5.7/en/server-options.html#option_mysqld_ignore-db-dir


luciddreamz:
@hhorak Please review and merge.

luciddreamz:
CC @bparees @oatmealraisin

PR #175, opened 2016-12-12
Added replica template to 5.7

oatmealraisin:
This is required for the origin tests, which pull mysql_replica.json. The only things changed are references to 5.6, which become 5.7

@bparees ptal?

### DIFF
--- a/5.7/examples/replica/README.md
+++ b/5.7/examples/replica/README.md
@@ -1,0 +1,151 @@
+# MySQL Replication Example
+
+**WARNING:**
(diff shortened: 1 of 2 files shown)


bparees:
can't we share a common mysql_replica.json and make the version a parameter?

bparees:
for now i'm going to merge it so we can move forward with getting the 5.7 imagestreams in.

oatmealraisin:
@bparees That works as well! On it

PR #180, opened 2017-04-11
add rhel6 5.7 build

tchughesiv:
some shell & python script changes required for compatibility w/ all 3 OS builds.

 rhel6 only supports MySQL 5.7 so far.

also made some yum & repo improvements for rhel7 build.

### DIFF
--- a/5.7/Dockerfile.rhel6
+++ b/5.7/Dockerfile.rhel6
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/rhel6
+FROM rhel6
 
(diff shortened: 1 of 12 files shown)


hhorak:
Hey, small update here, in order to avoid maintenance hell we're working on a solution for de-duplicating files, so it should be eventually possible to generate files that are different a bit across the versions and platforms, like Dockerfiles. That way we hopefully will be able to include rhel-6 variant at some point as well. Please, be patient.

tchughesiv:
Sounds great, tks for the update.

PR #184, opened 2017-05-24
Update Dockerfiles

omron93:
Use COPY instruction instead of ADD
(COPY is preffered for simple usage - see https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy)

Provide same labels for all base image variants (CentOS, RHEL, Fedora)
    - add name and version labels
    - remove architecture label for RHEL based images (this label is set in RHE$

(don't use separate instruction for each label)

@hhorak @praiskup @pkubatrh Please take a look and merge.


pkubatrh:
LGTM, merging.

PR #185, opened 2017-07-06
Load ini SQL files from /docker-entrypoint-initdb.d if any (for 5.6 in a first time)

mguillem:
This change adds the possibility to provide ini SQL files in /docker-entrypoint-initdb.d (only for 5.6 in a first time). This is really useful to provide images with an initial state for a test or QA environment.

The patch is deeply inspired from the official mysql docker image. Supporting exactly the same startup features would be really great. See "Initializing a fresh instance" in https://hub.docker.com/_/mysql/
or here for the sources
https://github.com/docker-library/mysql/blame/dc60c4b80f3eb5b7ef8b9ae09f16f6fab7a2fbf5/8.0/docker-entrypoint.sh#L177


hhorak:
@mguillem, thanks for the PR, we've implemented this a bit more generally, but your use case should be supported now, after 6b5c2858b938 has been implemented.

PR #188, opened 2017-07-14
Use direct link to Red Hat Container Catalog and RHEL-based image as example

hhorak:
SSIA


hhorak:
I hope no review needed here, it's only a slight change in the Doc.

PR #190, opened 2017-07-31
Move to newest common scripts

pkubatrh:
Remove help.1 files as they are now created before the image build
VERSION no longer used in make


omron93:
LGTM + tests are passing. So merging.

Thanks.

PR #191, opened 2017-08-11
Do not set innodb_additional_mem_pool_size which is deprecated

hhorak:
Option `innodb_additional_mem_pool_size` has been deprecated since version 5.6 and also it has no effect unless `innodb_use_sys_malloc` is turned off, which is not in our case.

See more at https://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_additional_mem_pool_size


hhorak:
@bparees I believe the option `innodb_additional_mem_pool_size` had never any real effect, but will rather wait to see whether you don't remember otherwise..

bparees:
@hhorak nope, can't say i recall anything about that setting behavior.

PR #195, opened 2017-09-01
Allow MySQL usernames up to 32 chars

mhutter:
Starting with MySQL 5.7, usernames can be up to 32 characters long.

Reference: https://dev.mysql.com/doc/refman/5.7/en/user-names.html

### DIFF
--- a/5.6/test/run
+++ b/5.6/test/run
@@ -224,9 +224,10 @@ function run_container_creation_tests() {
   try_image_invalid_combinations  -e MYSQL_ROOT_PASSWORD=root_pass
 
   local VERY_LONG_DB_NAME="very_long_database_name_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
+  local VERY_LONG_USER_NAME="very_long_user_name_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
   assert_container_creation_fails -e MYSQL_USER=user -e MYSQL_PASSWORD=pass
   assert_container_creation_fails -e MYSQL_USER=\$invalid -e MYSQL_PASSWORD=pass -e MYSQL_DATABASE=db -e MYSQL_ROOT_PASSWORD=root_pass
-  assert_container_creation_fails -e MYSQL_USER=very_long_username -e MYSQL_PASSWORD=pass -e MYSQL_DATABASE=db -e MYSQL_ROOT_PASSWORD=root_pass
+  assert_container_creation_fails -e MYSQL_USER=$VERY_LONG_USER_NAME -e MYSQL_PASSWORD=pass -e MYSQL_DATABASE=db -e MYSQL_ROOT_PASSWORD=root_pass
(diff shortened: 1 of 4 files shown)


omron93:
@mhutter Thank you for this improvement.

We try to have same scripts for all supported versions of mysql-container (`5.6` and `5.7` directories). So only Dockerfiles differs.

If some difference is needed (as in case of this PR), conditional is used - see for example https://github.com/sclorg/mysql-container/blob/2467c0f14869de2b2ce4213ea91f8bd33739e387/5.7/root/usr/share/container-scripts/mysql/common.sh#L93

So you can for example set `MAX_USERNAME_LENGTH` env. variable according to `MYSQL_VERSION` and use this new variable in `validate_variables()` function.

mhutter:
Hi @omron93, thanks for the feed back. I have updated the PR accordingly.

omron93:
[test]

omron93:
@mhutter Thanks.
Please, also tests have to updated in https://github.com/sclorg/mysql-container/blob/master/5.6/test/run#L229 (`very_long_username` is not long enough for 5.7)
So changes in way similar to adding `VERY_LONG_DB_NAME` variable is best approach.

mhutter:
@omron93 good point! I updated the test accordingly.

omron93:
[test]

5.6/test/run
```diff
@@ -10,7 +10,7 @@ set -o errexit
 set -o nounset
 shopt -s nullglob
 
-IMAGE_NAME=${IMAGE_NAME-centos/mysql-56-centos7-candidate}
+IMAGE_NAME=${IMAGE_NAME-centos/mysql-57-centos7-candidate}
```
omron93: This is wrong. I see you tried to have same test file for 5.6 and 5.7. +1

To do this properly you have available `$VERSION` variable. So this line could be

`IMAGE_NAME=${IMAGE_NAME-centos/mysql-${VERSION//.}-centos7-candidate}`

5.6/test/run
```diff
@@ -362,7 +363,7 @@ function run_tests() {
   create_container $name $envs
   test_connection "$name" "$USER" "$PASS"
   echo "  Testing scl usage"
-  test_scl_usage $name 'mysql --version' '5.6'
+  test_scl_usage $name 'mysql --version' '5.7'
```
omron93: Same as for image name.

`test_scl_usage $name 'mysql --version' "${VERSION}"`

mhutter:
@omron93 Whops, that totally slipped me...

I've fixed the tests and made sure both 5.6 and 5.7 use the same files.

omron93:
[test]

5.7/test/run
```diff
@@ -110,7 +110,7 @@ function create_container() {
 
 function run_change_password_test() {
   local tmpdir=$(mktemp -d)
-  chmod -R a+rwx "${tmpdir}"
+  mkdir "${tmpdir}/data" && chmod -R a+rwx "${tmpdir}"
```
omron93: @mhutter I hope last point - Please remove `mkdir "${tmpdir}/data" && ` (also for `5.6`)

`tmpdir` is mounted directly into `/var/lib/mysql/data` so with `data` dir in `tmpdir` mysql data dir is not empty. That's why `5.7` tests are failing.

mhutter: @omron93 ✅

hhorak:
[test]

PR #201, opened 2017-10-22
Implement s2i and general extendability support

hhorak:
This is very similar to the MariaDB: https://github.com/sclorg/mariadb-container/pull/45/

It adds extending support using [source-to-image](https://github.com/openshift/source-to-image).
    
For example, to build a customized MariaDB database image `my-mysql-rhel7`
with a configuration in `~/image-configuration/` run:

```
$ s2i build ~/image-configuration/ rhscl/mysql-56-rhel7 my-mysql-rhel7
```

The directory passed to `s2i build` can contain these directories:

`mysql-cfg/`
    When starting the container, files from this directory will be used as a configuration for the `mysqld` daemon. `envsubst` command is run on this file to still allow customization of the image using environmental variables

`mysql-pre-init/`
    Shell scripts (`*.sh`) available in this directory are sourced before `mysqld` daemon is started.

`mysql-init/`
    Shell scripts (`*.sh`) available in this directory are sourced when `mysqld` daemon is started locally. In this phase, use `${mysql_flags}` to connect to the locally running daemon, for example `mysql $mysql_flags < dump.sql`

Variables that can be used in the scripts provided to s2i:

`$mysql_flags`
    arguments for the `mysql` tool that will connect to the locally running `mysqld` during initialization

`$MYSQL_RUNNING_AS_MASTER`
    variable defined when the container is run with `run-mysqld-master` command

`$MYSQL_RUNNING_AS_SLAVE`
    variable defined when the container is run with `run-mysqld-slave` command

`$MYSQL_DATADIR_FIRST_INIT`
    variable defined when the container was initialized from the empty data dir

During `s2i build` all provided files are copied into `/opt/app-root/src` directory into the resulting image. If some configuration files are present in the destination directory, files with the same name are overwritten. Also only one file with the same name can be used for customization and user provided files are preferred over default files in `/usr/share/container-scripts/mysql/`- so it is possible to overwrite them.

Same configuration directory structure can be used to customize the image every time the image is started using `docker run`. The directory has to be mounted into `/opt/app-root/src/` in the image (`-v ./image-configuration/:/opt/app-root/src/`).
This overwrites customization built into the image.


hhorak:
This is the same change as was already reviewed in case of MariaDB, so I believe no extra review is needed. Tests passed, so merging.

PR #202, opened 2017-10-23
bz1505037: Add debugging info into passwd-change script

hhorak:
This is a different approach to fix lack of debugging information in passwd-change script, which was originally reported in https://bugzilla.redhat.com/show_bug.cgi?id=1505037 and #199.

### DIFF
--- a/root-common/usr/share/container-scripts/mysql/helpers.sh
+++ b/root-common/usr/share/container-scripts/mysql/helpers.sh
@@ -7,6 +7,14 @@ function log_and_run {
   "$@"
 }
 
+function log_debug {
+  CONTAINER_DEBUG=${CONTAINER_DEBUG:-}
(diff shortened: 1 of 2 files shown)


hhorak:
[test-openshift]

hhorak:
[test-openshift]

hhorak:
The function `log_debug`, `log_info`, and similar (we may also have `log_warning`) should be part of s2i-base.

root-common/usr/share/container-scripts/mysql/helpers.sh
```diff
@@ -7,6 +7,14 @@ function log_and_run {
   "$@"
 }
 
+function log_debug {
+  CONTAINER_DEBUG=${CONTAINER_DEBUG:-}
```
pkubatrh: Is this documented somewhere?

pkubatrh:
@hhorak ping

PR #206, opened 2017-11-16
Fix bogus file permissions

hhorak:


omron93:
LGTM + tests are passing.

Merging. Thanks.

PR #215, opened 2018-02-02
Add logic for upgrading datadir

hhorak:
A very same PR as https://github.com/sclorg/mariadb-container/pull/34, so probably doesn't require a special review, let's only consult it in https://github.com/sclorg/mariadb-container/pull/34.

### DIFF
--- a/root-common/usr/bin/run-mysqld-slave
+++ b/root-common/usr/bin/run-mysqld-slave
@@ -14,6 +14,11 @@ export_setting_variables
 
 log_volume_info $MYSQL_DATADIR
 
+# Just run normal server if the data directory is already initialized
+if [ -d "${MYSQL_DATADIR}/mysql" ]; then
+  exec /usr/bin/run-mysqld "$@"
+fi
+
 export MYSQL_RUNNING_AS_SLAVE=1
 
 # Generate the unique 'server-id' for this master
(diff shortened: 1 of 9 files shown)


hhorak:
[test-openshift]

hhorak:
[test-openshift]

hhorak:
[test-openshift]

hhorak:
[test-openshift]

pkubatrh:
[test-openshift]

pkubatrh:
@hhorak can you update common scripts to get more waiting time for docker registry?

hhorak:
[test-openshift]

hhorak:
[test-openshift]

hhorak:
[test-openshift]

hhorak:
Review was done as part of: https://github.com/sclorg/mariadb-container/pull/34, this is effectively same PR, so tests passing should be good for the review.

root-common/usr/bin/run-mysqld-slave
```diff
@@ -14,6 +14,11 @@ export_setting_variables
 
 log_volume_info $MYSQL_DATADIR
 
+# Just run normal server if the data directory is already initialized
+if [ -d "${MYSQL_DATADIR}/mysql" ]; then
```
gabemontero: @hhorak - turns out this change is breaking the mysql replica tests in https://github.com/openshift/origin/blob/master/test/extended/image_ecosystem/mysql_replica.go ... we've had to comment out that  test since this change dropped.

In our test, we force a restart of the slave pod by deleting the existing slave pod.  On the pod restart, this directory is discovered and we run `run-mysqld`, without setting `MYSQL_RUNNING_AS_SLAVE`.  As such, the process then attempts to validate various env vars in the `20-validate-variables.sh` script, and the slave restart fails.

How would you propose that slave restarts work when a PV is present with this change?

@bparees

PR #223, opened 2018-04-19
Do not use VOLUME command in Dockerfiles

pkubatrh:
As it does not work in Openshift Online for builder images for now.

### DIFF
--- a/5.6/Dockerfile
+++ b/5.6/Dockerfile
@@ -70,7 +70,9 @@ RUN rm -rf /etc/my.cnf.d/* && \
     /usr/libexec/container-setup && \
     rpm-file-permissions
 
-VOLUME ["/var/lib/mysql/data"]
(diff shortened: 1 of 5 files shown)


pkubatrh:
[test-openshift]

pkubatrh:
[test][test-openshift]

mfojtik:
@pkubatrh @hhorak this is breaking the new-app flow where we detected the VOLUME and notified users to create PVC for that volume so they data will be persisted.

hhorak:
Ok, will revert the change.

pkubatrh:
Since we triggered some failure on origin side of things, I now wonder if it was ok for us to remove the VOLUME definitions in other images...

hhorak:
@mfojtik thinks it is, because there is still the bz#1481918.. so we should probably have volume for databases, but not elsewhere.

PR #224, opened 2018-04-23
[DO NOT MERGE] Revert "Do not use VOLUME command in Dockerfiles"

hhorak:
This reverts commit 987e5c5e4e7cda060246e5d0007f2dca26516ab3, because it breaks things: https://github.com/openshift/origin/pull/19470

### DIFF
--- a/5.6/Dockerfile
+++ b/5.6/Dockerfile
@@ -70,9 +70,7 @@ RUN rm -rf /etc/my.cnf.d/* && \
     /usr/libexec/container-setup && \
     rpm-file-permissions
 
-# Not using VOLUME statement since it's not working in OpenShift Online:
-# https://github.com/sclorg/httpd-container/issues/30
-# VOLUME ["/var/lib/mysql/data"]
+VOLUME ["/var/lib/mysql/data"]
 
 USER 27
 
--- a/5.6/Dockerfile.rhel7
+++ b/5.6/Dockerfile.rhel7
@@ -70,9 +70,7 @@ RUN rm -rf /etc/my.cnf.d/* && \
     /usr/libexec/container-setup && \
     rpm-file-permissions
 
-# Not using VOLUME statement since it's not working in OpenShift Online:
-# https://github.com/sclorg/httpd-container/issues/30
-# VOLUME ["/var/lib/mysql/data"]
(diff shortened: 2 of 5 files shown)


hhorak:
[test-openshift]

hhorak:
This effectively reverts #223, but @mfojtik said we should wait for @bparees before merging. Ben, let us know :)

bparees:
https://github.com/sclorg/httpd-container/issues/30 hasn't been resolved, so why would we do this?

bparees:
since origin was fixed by https://github.com/openshift/origin/pull/19470, i would say nothing further needs to be done.

(other than perhaps introducing a new volume test in origin)

hhorak:
> sclorg/httpd-container#30 hasn't been resolved, so why would we do this?

It actually was fixed already, just github issues was not closed (I've done it now).

pkubatrh:
I thought we wanted to keep the issue open in order to track it for a future fix when Openshift Online is able to work with VOLUMEs

hhorak:
@pkubatrh ah, forgot about that, re-opening.

hhorak:
@bparees, Now I'm puzzled a bit, from what @mfojtik said I had a feeling that VOLUME set for databases images is actually beneficial, because users can be advised to use PV.. is that not a fair assumption? it seems to me like it's a trade-off between this advice and making s2i working in OpenShift online.. and I'm kinda clue-less what is more important use case. :)

hhorak:
@bparees Ok, thanks for you PoV. We did the change deliberately just in one database image at this point, to see what issues we can cause (we hoped for none, but were proofed wrong). I'd wait for further feedback (~month) before doing the same thing again for other images, so we don't break everything at once.
@mfojtik, if there are no other issues and we'll decide removing the VOLUME in other database images, can we expect the same thing fail again on your site?

bparees:
> @mfojtik, if there are no other issues and we'll decide removing the VOLUME in other database images, can we expect the same thing fail again on your site?

we'd have to scan our tests to see if we have other tests consuming the images you're changing and that expect to see volumes.  I wouldn't expect a bunch, we probably just picked the mysql image to exercise new-app.

hhorak:
It seems like we're ok to keep the image as it is, without VOLUME. So, closing.

PR #225, opened 2018-05-14
Deprecate MySQL 5.6

pkubatrh:


omron93:
@pkubatrh Shouldn't `5.6` be also remove from templates in `./examples/`? (MYSQL_VERSION variable - `5.5` is still there too)
Or these templates are not used?

pkubatrh:
Good point, thanks! Removed both versions from the description

pkubatrh:
[test]

pkubatrh:
[test]

omron93:
Thanks.

PR #228, opened 2018-06-05
Make mysql 5.6 EOL, freeze code

hhorak:
This just makes the mysql 5.6 code frozen, so we don't loose it. Should not touch 5.7 version at all.


hhorak:
@kubco2 Could you, please, take a look whether there are no obvious issues, like forgetting something?

PR #230, opened 2018-06-15
Revert part of "Sync with mysql a bit to make the comparisson simple"

### LINKED ISSUE #226: Regression introduce with restarting slave pods with PVs
See https://github.com/sclorg/mysql-container/pull/215/files#r190365336

Should the example/replica template even have a PVC for the slave?
If so, how do we refactor the reference PR/commit to allow slave pod restart per the openshift/origin extended tests I referenced?

thanks

hhorak:
This reverts part of the commit ff1a1179c1f7914ad4a1c3a60d58f4cf7c07cd0c.
This should hopefully fix #226.

### DIFF
--- a/root-common/usr/bin/run-mysqld-slave
+++ b/root-common/usr/bin/run-mysqld-slave
@@ -14,11 +14,6 @@ export_setting_variables
 
 log_volume_info $MYSQL_DATADIR
 
-# Just run normal server if the data directory is already initialized
-if [ -d "${MYSQL_DATADIR}/mysql" ]; then
-  exec /usr/bin/run-mysqld "$@"
-fi
-
 export MYSQL_RUNNING_AS_SLAVE=1
 
 # Generate the unique 'server-id' for this master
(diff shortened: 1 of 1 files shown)


hhorak:
[test-openshift]

hhorak:
Since it is a revert, I believe no review is needed.

bparees:
@hhorak what's your process for building/publishing the centos image these days?  (when should i expect to see an updated image w/ this change)

hhorak:
@bparees it should be built after each commit. so it's a matter of minutes usually.

bparees:
this has actually made things worse, the replicas aren't even coming up anymore:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/20012/test_pull_request_origin_extended_image_ecosystem/896/

bparees:
it looks like this test is failing because this check doesn't work anymore:

```
 mysqladmin -h localhost -uroot ping
mysqladmin: connect to server at 'localhost' failed
error: 'Your password has expired. To log in you must change it using a client that supports expired passwords.'
```

which is done by the test as part of prepping the replicas.  I need to look into whether that failure is expected.

bparees:
poking around I *think* this is an issue of the test case having bit-rotted even more than we realized rather than this PR causing problems.  I'll dig further and report back.  (I still think this reversion was needed though)

bparees:
ok i got the test passing again using the currently published mysql image.

Additional changes were required in the TC: https://github.com/openshift/origin/pull/20021

PR #241, opened 2018-10-03
run: Use hostname -I

pkubatrh:
As this option does not depend on name resolutions and filters out link local addresses


omron93:
Makes sense, thanks.

PR #247, opened 2018-12-05
Add rhel8 image for mysql 8.0

pkubatrh:


hhorak:
Fedora test fails because of wrong settings of capability, which should not be done in %post, but in %files section using parameters, so rpm -V failure is expected. It is gonna be fixed via https://src.fedoraproject.org/rpms/community-mysql/pull-request/1. Merging.

PR #248, opened 2019-02-03
Increase memory in the test because 8.0 simply does not work with just 256M any more

hhorak:
Also removing cgroup-limits binary as it is already in s2i-core image.

### DIFF
--- a/test/run
+++ b/test/run
@@ -339,7 +339,7 @@ function run_configuration_tests() {
 
   container_name=dynamic_config_test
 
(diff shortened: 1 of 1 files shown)


hhorak:
Fedora build is failing because the setcap is still set wrongly in the current latest package -- once https://src.fedoraproject.org/rpms/community-mysql/c/7c7c8b37e07ecaa3b86e0a2742368823cc2e1862?branch=master is fixed, it should start passing again. Won't block this PR because of that.

PR #251, opened 2019-03-05
remove shadow-utils from rpm -V check on rhel8

pkubatrh:
because there are some file capabilities dropped in container


kubco2:
What about `rpm -V --nocaps` , from manpage:        --nocaps Don't verify file capabilities.

pkubatrh:
I expect we want to verify capabilities in general. Its just the shadow-utils package that we know has issues even in the base image that we should ignore (since we cannot do anything about it anyway)

pkubatrh:
[test][test-openshift]

hhorak:
LGTM

PR #252, opened 2019-03-12
PoC: initialization of datadir

kubco2:
This PR is to point idea. It does not update all entry points of container and all example templates for openshift.
Container side:
Idea is to add argument "init" into entry points of containers.
If this argument is used container exits after datadir initialization.

Openshift side:
Init containers are started before the pod app containers are started. They do not have probes so they have enough time to initialize datadir.
https://docs.openshift.com/container-platform/3.6/architecture/core_concepts/pods_and_services.html#pods-services-init-containers

### DIFF
--- a/imagestreams/mysql-centos7.json
+++ b/imagestreams/mysql-centos7.json
@@ -27,42 +27,6 @@
         }
       },
       {
-        "name": "5.5",
-        "annotations": {
-          "openshift.io/display-name": "MySQL 5.5",
-          "openshift.io/provider-display-name": "Red Hat, Inc.",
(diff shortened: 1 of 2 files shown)


hhorak:
I don't see any specific problem at this point, but it is a big enough change to think it through very well, to make sure it does not have any side effects.

pkubatrh:
I am not sure I understand the use case for this? From docs it seems like this is a way to wait for some outside services without time-outing the pod that needs said services to function.

Is this supposed to help out with cases where the pod gets restarted before it is ready? Why not just increase the probe timeouts instead?

kubco2:
[test]

PR #253, opened 2019-03-21
Wait while shutting down server

kubco2:
https://dev.mysql.com/doc/refman/5.7/en/mysqladmin.html


kubco2:
mysqladmin already have socket parameter and it is not enough, because it waits only until pid is removed, but mysqld still exists and logs messages.

kubco2:
[test]

kubco2:
[test-openshift]

kubco2:
[test]

PR #269, opened 2019-10-01
Fixes BZ#1653215 Operation ALTER USER failed

phracek:
This issue fixes BZ#1653215.

This commit fixes problem formatting MySQL commands
in 50-passwd-change.sh script.
Correct BASH syntax as an input to MySQL command is used.

### DIFF
--- a/root-common/usr/share/container-scripts/mysql/init/50-passwd-change.sh
+++ b/root-common/usr/share/container-scripts/mysql/init/50-passwd-change.sh
@@ -4,9 +4,22 @@ password_change() {
   # Set the password for MySQL user and root everytime this container is started.
   # This allows to change the password by editing the deployment configuration.
   if [[ -v MYSQL_USER && -v MYSQL_PASSWORD ]]; then
-mysql $mysql_flags <<EOSQL
-      ALTER USER '${MYSQL_USER}'@'%' IDENTIFIED BY '${MYSQL_PASSWORD}';
+    TMP_FILE=$(mktemp /tmp/mysql_userXXXXX)
+mysql $mysql_flags >> $TMP_FILE <<EOSQL
+      SELECT user FROM mysql.user;
 EOSQL
+    grep -Fxq "$MYSQL_USER" "$TMP_FILE" > /dev/null
+    RETCODE=$?
+    if [[ $RETCODE -eq 0 ]]; then
+  mysql $mysql_flags <<EOSQL
+        ALTER USER '${MYSQL_USER}'@'%' IDENTIFIED BY '${MYSQL_PASSWORD}';
+EOSQL
+    else
+      log_info "User ${MYSQL_USER} does not exist in database."
+    fi
+    rm "${TMP_FILE}"
+    unset TMP_FILE
+
   fi
 
   # The MYSQL_ROOT_PASSWORD is optional, therefore we need to either enable remote
--- a/root-common/usr/share/container-scripts/mysql/init/50-passwd-change.sh
+++ b/root-common/usr/share/container-scripts/mysql/init/50-passwd-change.sh
@@ -4,21 +4,16 @@ password_change() {
   # Set the password for MySQL user and root everytime this container is started.
   # This allows to change the password by editing the deployment configuration.
   if [[ -v MYSQL_USER && -v MYSQL_PASSWORD ]]; then
-    TMP_FILE=$(mktemp /tmp/mysql_userXXXXX)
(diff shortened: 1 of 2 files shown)


hhorak:
[test]

hhorak:
Petr, using `<<'EOF'` instead of `<<EOSQL` makes the variable in the heredoc not expanded, why do you think this should fix the issue?

phracek:
I got inspired by https://github.com/mysql/mysql-docker/commit/13507953098b9cb707418a216ec5ac9d84326221 and https://github.com/sclorg/postgresql-container/blob/master/src/root/usr/share/container-scripts/postgresql/start/set_passwords.sh#L14
but did a copy-paste error. The next commit will fix it.

hhorak:
[test]

phracek:
In original 50-password-change.sh was used `<<EOSQL` instead of `<<-EOSQL`.
The `<<-EOSQL` is used then all leading tab characters are stripped from input lines.
I have written it a bit so each line starts at the beginning of the line.

root-common/usr/share/container-scripts/mysql/init/50-passwd-change.sh
```diff
@@ -4,24 +4,27 @@ password_change() {
   # Set the password for MySQL user and root everytime this container is started.
   # This allows to change the password by editing the deployment configuration.
   if [[ -v MYSQL_USER && -v MYSQL_PASSWORD ]]; then
-mysql $mysql_flags <<EOSQL
-      ALTER USER '${MYSQL_USER}'@'%' IDENTIFIED BY '${MYSQL_PASSWORD}';
-EOSQL
+mysql $mysql_flags \
+<<EOF
+ALTER USER '${MYSQL_USER}'@'%' IDENTIFIED BY '${MYSQL_PASSWORD}';
```
praiskup: Is there some guard against sql injection?

phracek: I have updated PR. Of course, I can return back `EOSQL` but I need to store the output of SELECT.

phracek:
Now I have updated `50-passwd-change.sh` file.
Unfortunately for some reason `grep` command crashes. Any idea why?

phracek:
I have changed script `50-password-change.sh` so that the script now checks
if a user exists in `mysql.user` database or not.

Problem is that if the container is run by a set of commands:
```
docker run -it --name=mysqldata -v $TMP:/var/lib/mysql/data -e MYSQL_USER=nwe12 -e MYSQL_PASSWORD=abcd -e MYSQL_DATABASE=mydb -e MYSQL_ROOT_PASSWORD=abcd docker-registry.engineering.redhat.com/phracek/mysql-57-rhel7:5.7 bash
/usr/bin/run-mysqld
```
grep tracebacks with this error:
```bash
---> 11:42:23     Setting passwords ...
-rw------- 1 mysql mysql 47 Oct  7 11:42 /opt/app-root/src/mysql_userkdlgc
uid=27(mysql) gid=27(mysql) groups=27(mysql),0(root)
---> 11:42:23     nwe12
---> 11:42:23     /opt/app-root/src/mysql_userkdlgc
user
newuser
root
mysql.session
mysql.sys
root
/usr/share/container-scripts/mysql/init//50-passwd-change.sh: line 17: file: command not found

Status information:

Current dir: /var/lib/mysql/data/
Running threads: 0  Stack size: 262144
Current locks:
lock: 0x55f021481e20:
```

but if when I do the same steps manually. All works fine.
I think I have missed some important stuff in mysql-container.
Any idea, what is wrong?

root-common/usr/share/container-scripts/mysql/init/50-passwd-change.sh
phracek: EOSQL is back again.

phracek:
I have solved the issue now. The output of MySQL `SELECT` command is stored into 
environment variable.
I have executed several local tests whether the functionality works properly. In case the user exists in `mysql.user` the database then the password is updated. In case the user does not exist in `mysql.user` the database then in container log can be seen the message, that the user does not exist.

phracek:
[test]

phracek:
[test-openshift]

pkubatrh:
[test-openshift]

hhorak:
Thanks a lot for the PR. I've changed the check section and used SQL directly to see whether the user exists. Also a regression test is added.

hhorak:
[test]

PR #271, opened 2019-10-22
Mysql 80 centos7

samyunodos:


phracek:
@samyunodos Please do not file PR against sclorg organization. It seems, like you have some automation, which creates a bunch of PR.

PR #277, opened 2019-12-06
Deprecated version MySQL 5.7

phracek:
Updated files are:
* Makefile
* README.md
* imagestream/*
* 5.7/root/usr/share/container-scripts/mysql/README.md


phracek:
Question is if we want to deprecated also the version from scripts as well. https://github.com/sclorg/mysql-container/blob/master/root-common/usr/share/container-scripts/mysql/init/40-datadir-action.sh

pkubatrh:
LGTM, thanks!

PR #280, opened 2020-01-22
Add MYSQL_DEFAULT_AUTHENTICATION_PLUGIN option

hhorak:
MySQL 8.0 introduced 'caching_sha2_password' as its default authentication plugin.
It is faster and provides better security then the previous default authentication plugin.
However, not all software implements this algorithm, and client applications might report
issue like "The server requested authentication method".

An example of such error: https://bugzilla.redhat.com/show_bug.cgi?id=1793116

Solution:

The plugin can be changed by setting `MYSQL_DEFAULT_AUTHENTICATION_PLUGIN` environment variable,
which accepts values `caching_sha2_password` (the default one) or `mysql_native_password`.
To change the behaviour back to the same behavior as MySQL 5.7 (for client applications that do not
support `caching_sha2_password`), set the `MYSQL_DEFAULT_AUTHENTICATION_PLUGIN=mysql_native_password`.

We keep `caching_sha2_password` as default because it has been default for long time and we should
not just change the config default. Since we didn't have many reports, it looks like users don't have
that many problems with the new authentication plugin. We should still offer a way to change it though.

### DIFF
--- a/8.0/root/usr/share/container-scripts/mysql/README.md
+++ b/8.0/root/usr/share/container-scripts/mysql/README.md
@@ -114,6 +114,8 @@ The following environment variables influence the MySQL configuration file. They
 **`MYSQL_LOG_QUERIES_ENABLED (default: 0)`**  
        To enable query logging set this to `1`
 
+**`MYSQL_DEFAULT_AUTHENTICATION_PLUGIN (default: caching_sha2_password)`**  
+       Set default authentication plugin. Accepts values `mysql_native_password` or `caching_sha2_password`.
 
 You can also set the following mount points by passing the `-v /host:/container` flag to Docker.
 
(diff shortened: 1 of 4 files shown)


gabemontero:
@hhorak @phracek so is this sufficient for getting the cakephp templates to work?  I think it is based on what I'm seeing, but don't feel confident enough to assume this.

phracek:
@gabemontero I have checked CentOS stream docker pull centos/mysql-80-centos7:latest, released 16 hours ago, and `default_authentication_plugin` is already present here based on this PR.

```bash
docker pull centos/mysql-80-centos7:latest
docker run --rm -it centos/mysql-80-centos7 bash
cat /usr/share/container-scripts/mysql/20-default-authentication-plugin.cnf
```

Issue #274, opened 2019-10-30
Openshift Pod logs print "Got an error reading communication packets"


imranrazakhan:
in pods logs i am continuously seeing "Got an error reading communication packets" but everything is working as expected
MYSQL 5.7
```
2019-10-30T21:38:55.333670Z 362671 [Note] Got an error reading communication packets
2019-10-30T21:38:56.180151Z 362673 [Note] Got an error reading communication packets
2019-10-30T21:38:57.333473Z 362674 [Note] Got an error reading communication packets
2019-10-30T21:38:59.360179Z 362675 [Note] Got an error reading communication packets
2019-10-30T21:39:01.335256Z 362676 [Note] Got an error reading communication packets
2019-10-30T21:39:03.333497Z 362678 [Note] Got an error reading communication packets
2019-10-30T21:39:05.333605Z 362679 [Note] Got an error reading communication packets
2019-10-30T21:39:06.180080Z 362680 [Note] Got an error reading communication packets
2019-10-30T21:39:07.333535Z 362681 [Note] Got an error reading communication packets
2019-10-30T21:39:09.333675Z 362683 [Note] Got an error reading communication packets
2019-10-30T21:39:11.333500Z 362686 [Note] Got an error reading communication packets
2019-10-30T21:39:13.333547Z 362687 [Note] Got an error reading communication packets
2019-10-30T21:39:15.333605Z 362689 [Note] Got an error reading communication packets
2019-10-30T21:39:16.180061Z 362690 [Note] Got an error reading communication packets
2019-10-30T21:39:17.333555Z 362691 [Note] Got an error reading communication packets
2019-10-30T21:39:19.333496Z 362692 [Note] Got an error reading communication packets
2019-10-30T21:39:21.333538Z 362694 [Note] Got an error reading communication packets
2019-10-30T21:39:23.333611Z 362695 [Note] Got an error reading communication packets
```

yannickcornaille:
I experience exactly the same behavior. If I deactivate the OpenShift livenessProbe the log disappears, but it is not a solution.
I also tried to increase the livenessProbe timeout (timeoutSeconds) but I still get this log.

imranrazakhan:
I changed to use MySQL Official Docker and done following changes

Current:
      readinessProbe:
        tcpSocket:
          port: 3306
        initialDelaySeconds: 20
        periodSeconds: 2
        timeoutSeconds: 1

[...]

PR #285, opened 2020-05-12
Use mysqladmin for the liveness probe

hhorak:
Use mysqladmin for the liveness probe

When using just TCP ping for the port, mysqld daemon expects some data that do not come. With "set global log_error_verbosity=3;", a warning is printed to the errlog, like this:
Got an error reading communication packets
(every 10s, depending on the liveness probe delay)

This makes issues when tools analyze the errlog, because those innoncent connection errors cannot be distinguised from real connection issues.

With using mysqladmin, similar to the readiness probe, the warning should not be printed any more.

Fixes: rhbz#1767393
Upstream issue: https://github.com/sclorg/mysql-container/issues/274
Upstream fix: https://github.com/sclorg/mysql-container/pull/285

### DIFF
--- a/examples/mysql-ephemeral-template.json
+++ b/examples/mysql-ephemeral-template.json
@@ -126,14 +126,15 @@
                   "initialDelaySeconds": 5,
                   "exec": {
                     "command": [ "/bin/sh", "-i", "-c",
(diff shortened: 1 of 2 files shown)


hhorak:
[test-openshift]

hhorak:
This was auto-closed by accident, it is actually valid to be fixed.

PR #289, opened 2020-07-03
Use mysqladmin for the liveness probe

hhorak:
This is the same thing as https://github.com/sclorg/mariadb-container/pull/123.

When using just TCP ping for the port, mysqld daemon expects some data that do not come. With "set global log_error_verbosity=3;", a warning is printed to the errlog, like this:
Got an error reading communication packets
(every 10s, depending on the liveness probe delay)

This makes issues when tools analyze the errlog, because those innoncent connection errors cannot be distinguised from real connection issues.

With using mysqladmin, similar to the readiness probe, the warning should not be printed any more.

Related: rhbz#1767393
Upstream issue: https://github.com/sclorg/mysql-container/issues/274
Upstream fix: https://github.com/sclorg/mysql-container/pull/285

### DIFF
--- a/examples/mysql-ephemeral-template.json
+++ b/examples/mysql-ephemeral-template.json
@@ -126,14 +126,15 @@
                   "initialDelaySeconds": 5,
                   "exec": {
                     "command": [ "/bin/sh", "-i", "-c",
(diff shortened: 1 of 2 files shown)


hhorak:
This was created by mistake #285 is supposed to be merged, it was closed by accident.

hhorak:
Since #285 cannot be re-opened, this actually will sever for fixing the issue.

hhorak:
[test-openshift]

hhorak:
Review of the same change already done in sclorg/mariadb-container#123. Merging.

PR #290, opened 2020-07-03
Support MYSQL_CHARSET and MYSQL_COLLATION variables

hhorak:
This is similar to https://github.com/sclorg/mariadb-container/pull/125.


hhorak:
Review of the same change already done in sclorg/mariadb-container#125, merging after tests passed.

hhorak:
This actually fixes #274.

PR #311, opened 2022-03-21
use composite build-and-push-action

zmiklank:


phracek (approved):
LGTM. Thanks for the PR

PR #319, opened 2022-05-31
Use rhel86 compose

phracek:


phracek:
`RHEL8` failures are expected.

PR #321, opened 2022-06-13
Use OpenShift internal registry instead of external

phracek:
Nowadays, we use the external OpenShift registry for the testing image.
OpenShift 4 provides an internal OpenShift registry. Let's use it.


phracek:
Could not resolve host problem

[test-openshift]

phracek:
Could not resolve host problem

[test-openshift]

zmiklank (approved):
LGTM.

PR #327, opened 2022-09-09
Testing single versions for container and openshift tests

phracek:


phracek:
[test-all]

phracek:
RHEL9 failures are caused by wrong compose https://github.com/sclorg/mysql-container/blob/master/.github/workflows/container-tests.yml#L49

zmiklank (approved):

PR #336, opened 2023-02-28
Fix OpenShift 4 tests

phracek:
This pull request fixes:

- updates container-common-scripts that were a bit outdated
- adds `run_latest_imagestreams test` even if there are not planned new versions
- fixes `test_mysql_integration` and tests both templates.


hhorak:
LGTM.

PR #349, opened 2023-08-31
imagestreams/mysql-centos: Update for C9S

LorbusChris:
Also switch to using the C8S image from quay.io for EL8 instead of pulling from docker.io, avoiding rate-limits.


LorbusChris:
cc @phracek

phracek:
[test-openshift]

PR #351, opened 2023-10-25
Add a test in Dockerfile directly to check mysql version

phracek:
Add a test in Dockerfile directly to check mysql version

Otherwise when a wrong stream is installed, it's only found by scl test, which is too late in the process.


phracek:
The functionality is added to RUN section after 'container-setup' because of capabilities
https://github.com/sclorg/mysql-container/blob/master/root-common/usr/libexec/container-setup#L61

[test]

phracek:
Failed test for RHEL7 is fixed by [8fc9a43](https://github.com/sclorg/mysql-container/pull/351/commits/8fc9a43ce7e2c48b1b03c332b1645b2f4fa762d2)

[test]

zmiklank (approved):

PR #355, opened 2024-02-26
cXs imagestreams: use -elX suffix for tag names

zmiklank:
to alling with the imagestreams of other sclorg images


phracek (approved):
LGTM. Thanks for the pull request.

[test-openshift]

phracek:
[test-openshift]

PR #357, opened 2024-03-18
Do not use dead symlink in repository.

phracek:
This pull request removes dead link from this upstream repository.

During the syncing source from upstream/GitLab repositories, the dead links are not sync bu
Testing Farm because security.
Therefore GitLab tests are failing because run script is missing.

Creating symlink in Dockerfile's solved this problem.

Similar pull requests are here: https://github.com/sclorg/postgresql-container/pull/552
https://github.com/sclorg/mariadb-container/pull/226


phracek:
[test]

PR #364, opened 2024-05-27
deprecate c8s images, as EOL for c8s is 31th of May 2024

zmiklank:


phracek (approved):
LGTM. Although no description.

PR #381, opened 2024-11-05
Run first PyTest executed on our Internal OpenShift cluster

phracek:
Run first PyTest executed on our Internal OpenShift cluster
and later on, on our Shared Cluster.

Container-ci-suite does not support yet, switching between clusters.


phracek:
[test-all]

phracek:
Close

PR #391, opened 2025-03-20
Update Dockerfile.rhel10 to support bind9.18-utils instead of bind-utils

phracek:
This pull request modifies Dockerfile.rhel10 for 8.4 version so bind9.18-utils is used instead of bind-utils


phracek:
[test][test-openshift]

PR #392, opened 2025-03-27
Rename bind-utils -> bind9.18-utils

phracek:
Rename bind-utils -> bind9.18-utils
The package has been renamed for RHEL9


phracek:
Let's test only for information
[test-openshift][test-openshift-pytest]

PR #393, opened 2025-04-22
Switch Ubuntu 20.04 to Ubuntu latest

phracek:
Switch Ubuntu 20.04 to Ubuntu latest\

The current version has been retired


phracek:
Running the tests is not needed. The change is affected as soon as PR is merged.

PR #395, opened 2025-04-29
Large series of enhancements mirroring changes accepted to MariaDB container

FaramosCZ:
I recommend to review the changes commit by commit


FaramosCZ:
[test]

FaramosCZ:
[test]

FaramosCZ:
[test]

FaramosCZ:
[test]

FaramosCZ:
[test]

FaramosCZ:
I intentionally only targeted versions that are relevant.
rhel 7, centos 7 and c8s all reached EOL, all are excluded,
and all of them just waits for the next batch of removal of EOLed versions from our repos.

8.0/Dockerfile.c9s
```diff
@@ -11,52 +11,54 @@ FROM quay.io/sclorg/s2i-core-c9s:c9s
 #  * $MYSQL_ROOT_PASSWORD (Optional) - Password for the 'root' MySQL account
 
 ENV MYSQL_VERSION=8.0 \
+    MYSQL_SHORT_VERSION=80 \
```
frenzymadness, line 14: FTR: I've tried to find a way to calculate this value from the variable above it, so we don't have to do it manually, and we'll never forget to do it, but there is no direct way, and all workarounds I've found are ugly.
Distgen would solve this by the way…

8.0/Dockerfile.fedora
```diff
@@ -36,16 +36,12 @@ LABEL summary="${SUMMARY}" \
...
-# in the container, this hack will be removed by installing /usr/bin/python from RPM.
-RUN ln -s /usr/bin/python3 /usr/bin/python
-
 # This image must forever use UID 27 for mysql user so our volumes are safe in the future. This should *never* change.
 # Instead of relying on the DB server package, we will do the setup ourselves before any package is installed
 RUN /usr/sbin/groupadd -g 27 -o -r mysql && \
     /usr/sbin/useradd -M -N -g mysql -o -r -d ${HOME} -s /sbin/nologin -c "MySQL Server" -u 27 mysql && \
     test "$(id mysql)" = "uid=27(mysql) gid=27(mysql) groups=27(mysql)" && \
-    INSTALL_PKGS="rsync tar gettext hostname bind-utils groff-base policycoreutils" && \
+    INSTALL_PKGS="rsync tar gettext hostname bind-utils groff-base policycoreutils python-unversioned-command" && \
```
frenzymadness, line 44: Do we need unversioned Python in DB container image? I mean, if the DB itself needs it, it should depend on it on RPM level. Otherwise, the best practice is to use the versioned command to be sure you run the version you want every time.

frenzymadness (commented):
I've finished the review and I have only one question about python-unversioned-command.

8.0/Dockerfile.fedora
FaramosCZ: The DB itself does not need it.

I checked the codebase and I haven't found any other part of the container that would depend on the unversioned command.

Furthermore, the `python-unversioned-command` package is already part of the `s2i-core` image, which require `python3` which brings the `python-unversioned-command` as a weak dependency.

I updated the PR and removed the installation of the `python-unversioned-command` in this image.

8.0/Dockerfile.c9s
FaramosCZ: I've found a way to compute it:

```
# Standalone ENV call so this value can be re-used in the other ENVs
ENV MYSQL_VERSION=8.0

ENV MYSQL_SHORT_VERSION=${MYSQL_VERSION//./} \
    APP_DATA=/opt/app-root/src \
    HOME=/var/lib/mysql \
    NAME=mysql
```

I updated the PR and applied it to all of the relevant containerfiles.

---

As a side note:
when the initial definition is not in a standalone stage, it can't be re-used by the other ENVs which are part of the same stage.
This actually was a bug I introduced in the mariadb-container and your notice made me think about it, use podman inspect, find the bug, fix it and verify the fix.

FaramosCZ:
[test]

PR #397, opened 2025-05-16
[revert] Revert using bash parameter substitution, 'bash' is not guaranteed

FaramosCZ:
Build system without bash fails with:
 | Error: resolving step {Value:env Next:0xc0001524d0 Children:[]
 | Attributes:map[] Original:ENV MYSQL_SHORT_VERSION=${MYSQL_VERSION//./}
 | VERSION=${MYSQL_VERSION}     APP_DATA=/opt/app-root/src     HOME=/var/lib/mysql
 | NAME=mariadb     SUMMARY="MariaDB ${MYSQL_VERSION} SQL database server"
 | DESCRIPTION="..."
 | Flags:[] StartLine:16 EndLine:25}: Missing ':' in substitution: ${MYSQL_VERSION//./}

I tried a lot of ways to work around this, none worked.

Using subshells
 - ENV does not support white characters on the right side of assignment Using subshells with scrambled whitepsace characters
 - ENV won't invoke the subshell, will treat right side of the assignement as a simple text Using RUN to set $SHELL, trying to use 'export'
 - won't be preserved between layers using SHELL directive
 - Docker specific

Fedora only knows '/bin/bash' and 'sh' is just a symbolic link to '/bin/bash'. So the 'sh' unable to do the parameter expansions must come from the build host.

My guess would be that setting 'shell: bash' here:
  https://github.com/sclorg/build-and-push-action/blob/main/action.yml#L115
might help.

However this seems to be impossible to force inside of the containerfile itself, so I'm reverting the bash parameter substitution code entirely.

### DIFF
--- a/8.0/Dockerfile.c9s
+++ b/8.0/Dockerfile.c9s
@@ -10,10 +10,8 @@ FROM quay.io/sclorg/s2i-core-c9s:c9s
 #  * $MYSQL_DATABASE - Name of the database to create
 #  * $MYSQL_ROOT_PASSWORD (Optional) - Password for the 'root' MySQL account
 
-# Standalone ENV call so this value can be re-used in the other ENVs
-ENV MYSQL_VERSION=8.0
-
-ENV MYSQL_SHORT_VERSION=${MYSQL_VERSION//./} \
+ENV MYSQL_VERSION=8.0 \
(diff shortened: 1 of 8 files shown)


FaramosCZ:
[test]

phracek (approved):
LGTM

PR #399, opened 2025-05-22
[bugfix] The version must be set in a separate stage, so it can be reused in the other calls

FaramosCZ:
otherwise it's not available when the rest of the ENV block is processed and the ENV values relying on this variable are malformed.

This commit was mistakenly reverted as a part of revert of a different change


FaramosCZ:
[test]

FaramosCZ:
[test]

FaramosCZ:
[test]

phracek (approved):
LGTM.

PR #396, opened 2025-05-14
Allow testing RHEL10 imagestreams on OpenShift 4 and inside templates

phracek:
Allow testing RHEL10 imagestreams on OpenShift 4 and inside templates

Enable testing version 8.4 also for RHEL9

PR #400, opened 2025-05-29
Update GitHub action

phracek:
Disable auto-merge-on-demand

Add version 8.4 to OpenShift 4 tests


phracek:
The test is not needed. No code is changed. It will be fully tested by #396 .

PR #410, opened 2025-09-19
Remove RHEL7/CentOS7 and  scl_enable

phracek:
This removes RHEL7 / CentOS 7 and scl_enable from container itself.

RHEL7 / CentOS 7 reached GA.

The commit contains:
- Moving versions to the latest one as we can like rhel8/mysql-80 -> rhel9/mysql-80 or rhel9/mysql-84 -> rhel10/mysql-84
- Removing RHEL 7 / CentOS 7
- Updated README.md files
- updating templates to use the latest imagestreams


phracek:
Question is, shell I remove also https://github.com/sclorg/mysql-container/blob/master/test/run#L300 where is `/opt/rh`?

README.md
```diff
@@ -102,7 +102,7 @@ Test
 This repository also provides a test framework, which checks basic functionality
 of the MySQL image.
 
-Users can choose between testing MySQL based on a RHEL or CentOS image.
+Users can choose between testing MySQL based on a RHEL or CentOS Stream image.
 
 *  **RHEL based image**
```
pkhartsk, line 107: Maybe we should consider using rhel9/10 and mysql 8.4 here?

phracek: Thanks. The whole README.md file mentinoed as version `8.0` as `8.4`. The versions are mentioned here: https://github.com/sclorg/mysql-container/blob/master/README.md#versions.

Each version has its own README.md file e.g. https://github.com/sclorg/mysql-container/tree/master/8.4

test/run
```diff
@@ -550,9 +550,6 @@ function run_general_tests() {
 
 function get_previous_major_version() {
   case "${1}" in
-    5.5) echo "5.1" ;;
-    5.6) echo "5.5" ;;
-    5.7) echo "5.6" ;;
     8.0) echo "5.7" ;;
     8.4) echo "8.0" ;;
     10.0) echo "5.5" ;;
```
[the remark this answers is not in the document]
phracek: No I will remove it.

phracek:
> In the MR and commits, you mention removing c7s and RHEL-7. But you removed files for c8s as well, so if it's not a mistake, maybe add a comment why you did that as well

PR #437, opened 2026-05-28
add distgen support

ndavidova:


ndavidova:
[test]

ndavidova:
[test]

src/Dockerfile
```diff
@@ -0,0 +1,147 @@
...
+      maintainer="SoftwareCollections.org <sclorg@redhat.com>"
+{% endif %}
+
+EXPOSE 3306
+
+{% if spec.prod == 'fedora' %}# This image must forever use UID 27 for mysql user so our volumes are safe in the future. This should *never* change.
+# Instead of relying on the DB server package, we will do the setup ourselves before any package is installed
+{% else %}# This image must forever use UID 27 for mysql user so our volumes are
+# safe in the future. This should *never* change, the last test is there
+# to make sure of that.
```
pkhartsk: Seems useless to do this

pkhartsk:
Would it also make sense to include the `s2i-common` directories in the sources? Or just make them symlinks to the one already present in the project root

src/Dockerfile
pkhartsk: I'm aware that this would cause the dockerfiles to differ cosmetically, but functionally they remain the same, and this is just pretty much the same comment...

src/Dockerfile
```diff
@@ -0,0 +1,147 @@
...
+The mysqld server daemon accepts connections from clients and provides access to content from \
+MySQL databases on behalf of the clients."
+
+LABEL summary="${SUMMARY}" \
+      description="${DESCRIPTION}" \
+      io.k8s.description="${DESCRIPTION}" \
+      io.k8s.display-name="MySQL ${MYSQL_VERSION}" \
+      io.openshift.expose-services="3306:mysql" \
+      io.openshift.tags="database,${NAME},${NAME}${MYSQL_SHORT_VERSION}{% if spec.prod == 'fedora' and spec.version == '8.0' %},rh-${NAME}${MYSQL_SHORT_VERSION}{% elif spec.prod != 'fedora' %},${NAME}-${MYSQL_SHORT_VERSION}{% endif %}" \
+      com.redhat.component="{% if spec.prod == 'rhel8' or spec.prod == 'rhel9' or spec.prod == 'rhel10' or spec.prod == 'c9s' or spec.prod == 'c10s' %}${NAME}-${MYSQL_SHORT_VERSION}-container{% else %}${NAME}{% endif %}" \
```
pkhartsk: Can be simplified to `if spec.prod != 'fedora'`

src/Dockerfile
```diff
@@ -0,0 +1,147 @@
...
+      name="{{ spec.img_name }}" \
+{% if spec.prod == 'fedora' %}      version="${MYSQL_VERSION}" \
+{% elif spec.prod != 'rhel10' %}      version="1" \
```
pkhartsk, line 35: This makes no sense to me, why would the version be 1? Shouldn't it always be MYSQL_VERSION? Guessing this is a mistake in the .rhelx Dockerfiles?

src/Dockerfile
```diff
@@ -0,0 +1,147 @@
...
+{% endif %}{% if spec.license_terms %}      com.redhat.license_terms="{{ spec.license_terms }}" \
+{% endif %}      usage="{% if spec.prod == 'c9s' or spec.prod == 'c10s' %}podman run -d -e MYSQL_USER=user -e MYSQL_PASSWORD=pass -e MYSQL_DATABASE=db -p 3306:3306 quay.io/{{ spec.img_name }}:{{ spec.prod }}{% elif spec.prod == 'fedora' %}docker run -d -e MYSQL_USER=user -e MYSQL_PASSWORD=pass -e MYSQL_DATABASE=db -p 3306:3306 quay.io/{{ spec.img_name }}{% else %}podman run -d -e MYSQL_USER=user -e MYSQL_PASSWORD=pass -e MYSQL_DATABASE=db -p 3306:3306 {{ spec.img_name }}{% endif %}" \
```
pkhartsk, line 37: nit: these endif's at the starts of lines are ugly

src/Dockerfile
```diff
@@ -0,0 +1,147 @@
...
+    rpm -V ${INSTALL_PKGS} && \
+    dnf -y clean all --enablerepo='*' && \
+    mkdir -p ${HOME}/data && chown -R mysql:{{ spec.chown_group }} ${HOME} && \
+    test "$(id mysql)" = "uid=27(mysql) gid=27(mysql) groups=27(mysql)"
+{% elif spec.version == '8.0' %}RUN INSTALL_PKGS="{{ spec.pkgs }}" && \
+    yum install -y --setopt=tsflags=nodocs ${INSTALL_PKGS} && \
+    rpm -V ${INSTALL_PKGS} && \
+    yum -y clean all --enablerepo='*' && \
+    mkdir -p ${HOME}/data && chown -R mysql.0 ${HOME} && \
+    test "$(id mysql)" = "uid=27(mysql) gid=27(mysql) groups=27(mysql)"
```
pkhartsk, line 71: These modular installs seem very similar, maybe the if can be divided into something like
`if spec.prod in ['rhel8', ... 'c9s']` for modules where the module is enabled, and then a more common way to install the packages?

src/Dockerfile
```diff
@@ -0,0 +1,147 @@
...
+ENV CONTAINER_SCRIPTS_PATH=/usr/share/container-scripts/mysql \
+    MYSQL_PREFIX=/usr
+
+{% if spec.prod == 'rhel8' and spec.version == '8.0' %}COPY {{ spec.version }}/root-common /
+COPY {{ spec.version }}/s2i-common/bin/ $STI_SCRIPTS_PATH
+COPY {{ spec.version }}/root /
+{% else %}COPY ${MYSQL_VERSION}/root-common /
+COPY ${MYSQL_VERSION}/s2i-common/bin/ ${STI_SCRIPTS_PATH}
+COPY ${MYSQL_VERSION}/root /
+{% endif %}
```
pkhartsk: Aren't these the exact same?

src/Dockerfile
```diff
@@ -0,0 +1,147 @@
...
+
+# Hard links are not supported in Testing Farm approach during sync to guest
+# operation system. Therefore tests are failing on error
+# /usr/libexec/s2i/run no such file or directory
+{% if spec.prod == 'rhel8' and spec.version == '8.0' %}RUN ln -s /bin/run-mysqld $STI_SCRIPTS_PATH/run
+{% else %}RUN ln -s /bin/run-mysqld ${STI_SCRIPTS_PATH}/run
+{% endif %}
```
pkhartsk: Also the exact same.

src/Dockerfile
```diff
@@ -0,0 +1,147 @@
...
+# Also reset permissions of filesystem to default values
+{% if spec.prod == 'rhel8' and spec.version == '8.0' %}RUN rm -rf /etc/my.cnf.d/* && \
+    /usr/libexec/container-setup && \
+    rpm-file-permissions && \
+    /usr/libexec/mysqld -V | grep -qe "$MYSQL_VERSION\." && echo "Found VERSION $MYSQL_VERSION"
+{% else %}RUN rm -rf /etc/my.cnf.d/* && \
+    /usr/libexec/container-setup && \
+    rpm-file-permissions && \
+    /usr/libexec/mysqld -V | grep -qe "${MYSQL_VERSION}\." && echo "Found VERSION ${MYSQL_VERSION}"
+{% endif %}
```
pkhartsk: Also the exact same.

src/Dockerfile
```diff
@@ -0,0 +1,147 @@
...
+      com.redhat.component="{% if spec.prod == 'rhel8' or spec.prod == 'rhel9' or spec.prod == 'rhel10' or spec.prod == 'c9s' or spec.prod == 'c10s' %}${NAME}-${MYSQL_SHORT_VERSION}-container{% else %}${NAME}{% endif %}" \
+      name="{{ spec.img_name }}" \
+{% if spec.prod == 'fedora' %}      version="${MYSQL_VERSION}" \
+{% elif spec.prod != 'rhel10' %}      version="1" \
+{% endif %}{% if spec.license_terms %}      com.redhat.license_terms="{{ spec.license_terms }}" \
+{% endif %}      usage="{% if spec.prod == 'c9s' or spec.prod == 'c10s' %}podman run -d -e MYSQL_USER=user -e MYSQL_PASSWORD=pass -e MYSQL_DATABASE=db -p 3306:3306 quay.io/{{ spec.img_name }}:{{ spec.prod }}{% elif spec.prod == 'fedora' %}docker run -d -e MYSQL_USER=user -e MYSQL_PASSWORD=pass -e MYSQL_DATABASE=db -p 3306:3306 quay.io/{{ spec.img_name }}{% else %}podman run -d -e MYSQL_USER=user -e MYSQL_PASSWORD=pass -e MYSQL_DATABASE=db -p 3306:3306 {{ spec.img_name }}{% endif %}" \
+      maintainer="SoftwareCollections.org <sclorg@redhat.com>"
+{% endif %}
```
pkhartsk: Extremely long if and the only differences I see are the use of `{{spec.version}}` instead of `${MYSQL_VERSION}` and similar small things that evaluate to the same thing during buildtime. Maybe this can be simplified somehow?

pkhartsk (changes requested):
Found some possible simplifications of the (IMO) very complicated `src/Dockerfile`

ndavidova:
> Would it also make sense to include the `s2i-common` directories in the sources? Or just make them symlinks to the one already present in the project root

I believe this should already be present.

ndavidova:
[test]

PR #438, opened 2026-05-28
Test distgen-check workflow

ndavidova:
Testing if distgen-check workflow runs correctly when a PR is created.


ndavidova:
[test]

PR #441, opened 2026-06-23
OKD-380: Add missing namespace to image trigger annotations

pskrbasu:
## Problem

Deploying MySQL from the OKD 4.22 Software Catalog fails with:

```
spec.containers[0].image: Invalid value: " ": must not have leading or trailing whitespace
```

Reported in https://github.com/okd-project/okd/issues/2337

## Root Cause

When templates were migrated from `DeploymentConfig` to `Deployment`, the `image.openshift.io/triggers` annotation dropped the `namespace` field from the `ImageStreamTag` reference. Without it, the image trigger controller defaults to the Deployment's own namespace instead of `openshift`, so it can't find the ImageStream and the container image is never resolved.

## Fix

- Add `"namespace":"${NAMESPACE}"` to the trigger annotation's `from` object in both `mysql-persistent` and `mysql-ephemeral` templates
- Remove a stray `image.openshift.io/triggers` annotation from the Service object in `mysql-persistent-template.json` — it was a copy-paste error that referenced `${MARIADB_VERSION}` (Services don't have containers, so this annotation had no effect)

## Impact

These template files are the upstream source for `openshift/library` (via `make import`) and ultimately `openshift/cluster-samples-operator`. An immediate fix was applied directly to the operator in [openshift/cluster-samples-operator#701](https://github.com/openshift/cluster-samples-operator/pull/701). This upstream fix ensures the bug doesn't recur on the next sync.

## Summary by CodeRabbit

* **Chores**
  * Updated example MySQL deployment templates (ephemeral and persistent) to explicitly reference image streams with namespace parameters, ensuring proper deployment triggers across different environments.
  * Removed redundant image stream annotation from the Service resource in the persistent template to streamline configuration.

### DIFF
--- a/examples/mysql-ephemeral-template.json
+++ b/examples/mysql-ephemeral-template.json
@@ -75,7 +75,7 @@
         "name": "${DATABASE_SERVICE_NAME}",
         "annotations": {
           "template.alpha.openshift.io/wait-for-ready": "true",
-          "image.openshift.io/triggers": "[{\"from\":{\"kind\":\"ImageStreamTag\",\"name\":\"mysql:${MYSQL_VERSION}\"},\"fieldPath\": \"spec.template.spec.containers[0].image\"}]"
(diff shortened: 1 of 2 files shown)


pskrbasu:
[test-openshift-pytest]

phracek (approved):
LGTM

phracek:
[test-openshift-pytest]

